PR #63, opened 2020-01-02
feat: add key transactions

sanderblue:
Resolves: #20

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -1,0 +1,19 @@
+package utils
+
+import (
+	"strconv"
+	"strings"
+)
+
+// IntArrayToString converts an array of integers
+// to a comma-separated list string -
+// e.g [1, 2, 3] will be converted to "1,2,3".
+func IntArrayToString(integers []int) string {
+	sArray := []string{}
+
+	for _, n := range integers {
+		sArray = append(sArray, strconv.Itoa(n))
+	}
+
+	return strings.Join(sArray, ",")
+}
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -9,10 +9,10 @@ import (
 // to a comma-separated list string -
 // e.g [1, 2, 3] will be converted to "1,2,3".
 func IntArrayToString(integers []int) string {
-	sArray := []string{}
+	sArray := make([]string, len(integers))
 
-	for _, n := range integers {
-		sArray = append(sArray, strconv.Itoa(n))
+	for i, n := range integers {
+		sArray[i] = strconv.Itoa(n)
 	}
 
 	return strings.Join(sArray, ",")
--- a/internal/utils/utils_test.go
+++ b/internal/utils/utils_test.go
@@ -11,8 +11,17 @@ import (
 func TestIntArrayToString(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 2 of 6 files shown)


pkg/apm/key_transactions.go
```diff
@@ -0,0 +1,89 @@
...
+	if params.IDs != nil && len(params.IDs) > 0 {
+		paramsMap["filter[ids]"] = intArrayToString(params.IDs)
+	}
+
+	return paramsMap
+}
+
+// Converts an array of integers to a comma-separated list string.
+// Example: [1, 2, 3] will be converted to "1,2,3"
+func intArrayToString(integers []int) string {
```
sanderblue: TODO: potentially put this in a centralized utils package or in a package-scoped `utils.go` file.

sanderblue: Updated.

internal/utils/utils.go
```diff
@@ -0,0 +1,19 @@
...
+// e.g [1, 2, 3] will be converted to "1,2,3".
+func IntArrayToString(integers []int) string {
+	sArray := []string{}
+
+	for _, n := range integers {
+		sArray = append(sArray, strconv.Itoa(n))
+	}
+
+	return strings.Join(sArray, ",")
+}
```
jthurman42: Looping through append could be optimized since we know the length of `integers`:

```
func IntArrayToString(integers []int) string {
	sArray := make([]string, len(integers))

	for i, n := range integers {
		sArray[i] = strconv.Itoa(n)
	}

	return strings.Join(sArray, ",")
}
```

internal/utils/utils_test.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func TestIntArrayToString(t *testing.T) {
+	t.Parallel()
+
+	arr := []int{1, 2, 3, 4}
+	result := IntArrayToString(arr)
+
+	assert.Equal(t, "1,2,3,4", result)
+}
```
jthurman42: Should add tests for Empty/Single/Multiple value slices:

```
func TestIntArrayToString(t *testing.T) {
	t.Parallel()

	var result string

	// empty
	result = IntArrayToString([]int{})
	assert.Equal(t, "", result)

	// single
	result = IntArrayToString([]int{1})
	assert.Equal(t, "1", result)

	// multiple
	result = IntArrayToString([]int{1, 2, 3, 4})
	assert.Equal(t, "1,2,3,4", result)
}
```

internal/utils/utils.go
```diff
@@ -0,0 +1,19 @@
+package utils
```
jthurman42: nit: Utils is a great package name.  Should this file be called `strings` or something else more specific to the type of util it is?

sanderblue: Fair question. My thought was that this was to not overthink that part since it's only one util function for now. And maybe as things are added, the proper "homes" would present themselves. That being said, in this case, I would think the file name for this util function would be `arrays.go` since the input is an array/slice.

internal/utils/utils_test.go
sanderblue: Updated, thanks!

internal/utils/utils.go
sanderblue: Updated, thanks!

Issue #64, opened 2020-01-03
Infra Alert Conditions: support thresholds with float values


sanderblue:
### Objective
Infrastructure alert condition thresholds can apparently be set to float values via the New Relic UI. Our Go client should support this as well. The type update would happen [here](https://github.com/newrelic/newrelic-client-go/blob/master/pkg/alerts/infrastructure_conditions_types.go#L30).

The original [issue](https://github.com/terraform-providers/terraform-provider-newrelic/issues/291) was reported in the `terraform-provider-newrelic` project.

### Acceptance Criteria
- Infrastructure alert condition thresholds support using float values.

jthurman42:
Resolved in #71

PR #71, opened 2020-01-06
bug: Change Infra alert condition value to float

jthurman42:
The API endpoint accepts a float for Infrastructure Alert Condition value, update the client to use float.

* Change field from `int` to `float64`
* Add test for Warning threshold
* Test using float and int in JSON response body


jthurman42:
Resolves #64

Issue #54, opened 2019-12-23
Ensure consistency among all resource endpoints


sanderblue:
### Objective
Provide a consistent contract for all methods.

### Acceptance Criteria
- Favor slices of pointers over slices of structs for endpoints that return collections. 
- Favor multiple return values for all endpoints, always returning the object(s) being operated on.
    - e.g. always return a tuple `func UpdateEntity(entity Entity) (*Entity, error) {}` even if the endpoint this method uses doesn't really return us a response body

PR #82, opened 2020-01-08
refactor: API and code consistency updates

sanderblue:
Resolves: #54


pkg/dashboards/dashboards_test.go
```diff
@@ -288,34 +287,24 @@ var (
 
 func TestListDashboards(t *testing.T) {
 	t.Parallel()
-	apm := NewTestDashboards(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	client := NewTestDashboards(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
jthurman42: This should use the `newMockResponse`

sanderblue: Yup, I'll get this consolidated in a fast-follow PR

PR #84, opened 2020-01-09
Auto-generation of CHANGELOG

jthurman42:
This adds `git-chglog` as a build dependency and automatically execs it on:

* `make changelog`

Not automatically run on build to avoid merge conflicts...  Should be come part of the release cycle once one is created.


.chglog/config.yml
```diff
@@ -0,0 +1,38 @@
...
+  commit_groups:
+    title_maps:
+      docs: Documentation Updates
+      feat: Features
+      fix: Bug Fixes
+
+  refs:
+    actions:
+      - Closes
+      - Fixes
```
sanderblue: Can we add "Resolves"?

PR #85, opened 2020-01-09
feat: introduce logging

### LINKED ISSUE #87: Add logging
AC:
* Add logging support to the client

ctrombley:
Resolved #87. 

This PR brings logging to the client via the logrus package.


internal/http/client.go
```diff
@@ -84,11 +83,15 @@ func NewClient(config config.Config) NewRelicClient {
 
 		level, err := log.ParseLevel(config.LogLevel)
 		if err != nil {
-			panic(err)
+			log.Warn(fmt.Sprintf("could not parse log level '%s', logging will proceed at info level", config.LogLevel))
+			level = log.InfoLevel
```
jthurman42: This should output `defaultLogLevel` and parse that level again since we have a constant, easy to miss this if we change the default level later.

internal/logging/default_field_hook.go
```diff
@@ -0,0 +1,20 @@
...
+type DefaultFieldHook struct{}
+
+// Levels determines which levels will get default fields.
+func (h *DefaultFieldHook) Levels() []log.Level {
+	return log.AllLevels
+}
+
+// Fire is the hook for each log message.
+func (h *DefaultFieldHook) Fire(e *log.Entry) error {
+	e.Data["newrelic-client-go"] = version.Version
```
jthurman42: Nice

internal/http/client.go
```diff
@@ -62,11 +67,35 @@ func NewClient(config config.Config) NewRelicClient {
...
+			level, _ = log.ParseLevel(defaultLogLevel)
+		}
+
+		log.SetLevel(level)
+
+		if config.LogJSON {
+			log.SetFormatter(&log.JSONFormatter{})
+		}
+	}
+
```
jthurman42: It feels like we should be able to just do `r.Logger = config.Logger` and handle creating a logger within Config.

pkg/config/config.go
```diff
@@ -39,4 +39,18 @@ type Config struct {
...
+	LogJSON  bool
+	Logger   Logger
+}
+
+// Logger interface implements a simple logger.
+type Logger interface {
+	Error(string, ...interface{})
+	Info(string, ...interface{})
+	Debug(string, ...interface{})
+	Warn(string, ...interface{})
```
jthurman42: We should move this into the `logging` package and reference it here instead.

pkg/config/config.go
```diff
@@ -39,4 +39,18 @@ type Config struct {
 	HTTPTransport *http.RoundTripper
 	UserAgent     string
 	Region        RegionType
+
+	// LogLevel can be one of the following values:
+	// "panic", "fatal", "error", "warn", "info", "debug", "trace"
+	LogLevel string
+	LogJSON  bool
+	Logger   Logger
```
jthurman42: Should this be a nested config struct? `Config.Logging.*`

internal/http/client.go
jthurman42: `logging` should handle creation of the default logger.

```
if config.Logger != nil {
  // same
} else {
  r.Logger = logging.DefaultLogger
  // or
  // r.Logger = logging.New(config)
  // or if moving logging into a deeper struct in config:
  // r.Logger = logging.New(config.Logger)
}
```

pkg/config/config.go
ctrombley: Logging is internal, I had moved it here so it was more discoverable for folks implementing it.

pkg/alerts/alerts.go
```diff
@@ -24,9 +27,17 @@ func New(config config.Config) Alerts {
...
+		config.Logger = logging.NewStructuredLogger().
+			SetDefaultFields(map[string]string{"newrelic-client-go": version.Version}).
+			LogJSON(config.LogJSON).
+			SetLogLevel(config.LogLevel)
+	}
+
 	pkg := Alerts{
 		client:      http.NewClient(config),
 		infraClient: infraClient,
+		logger: config.Logger,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		logger:      config.Logger,
```

pkg/config/config.go
```diff
@@ -39,4 +42,25 @@ type Config struct {
...
+	LogLevel string
+	LogJSON  bool
+	Logger   logging.Logger
+}
+
+// GetLogger returns a logger instance based on the config values.
+func (c *Config) GetLogger() logging.Logger {
+	if c.Logger != nil {
+		return c.Logger
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

PR #88, opened 2020-01-10
Added tests, and possible Mock refactor for discussion

jthurman42:
It's not as clean as I'd like, but a starting point for refactoring the Mock setup for unit testing.

Also includes a test for the Version package to make sure we're keeping it consistent with the latest tagged release.


jthurman42:
Converted most of Alerts to use the central mock, now the rest of the packages need to be converted.

PR #89, opened 2020-01-13
fix: make use of ErrorNotFound type for Get methods that are based on List methods

ctrombley:
The `ErrorNotFound` type is meaningful for downstream use in the Terraform provider.  It initially represented a 404 response from a REST Show endpoint, but it should also be used for Get methods that do not have corresponding Show endpoints and need to iterate over results returned from a List endpoint .

This PR puts the ErrorNotFound type in place for these scenarios, and also allows for custom error messages.

### DIFF
--- a/pkg/alerts/channels.go
+++ b/pkg/alerts/channels.go
@@ -2,6 +2,8 @@ package alerts
 
 import (
 	"fmt"
+
+	"github.com/newrelic/newrelic-client-go/pkg/errors"
 )
 
 // ListChannels returns all alert channels for a given account.
@@ -38,8 +40,9 @@ func (alerts *Alerts) GetChannel(id int) (*Channel, error) {
 			return channel, nil
 		}
 	}
-
-	return nil, fmt.Errorf("no channel found for id %d", id)
+	return nil, &errors.ErrorNotFound{
+		Message: fmt.Sprintf("no channel found for id %d", id),
+	}
 }
 
 // CreateChannel creates an alert channel within a given account.
--- a/pkg/alerts/channels.go
+++ b/pkg/alerts/channels.go
@@ -40,9 +40,7 @@ func (alerts *Alerts) GetChannel(id int) (*Channel, error) {
 			return channel, nil
 		}
 	}
-	return nil, &errors.ErrorNotFound{
-		Message: fmt.Sprintf("no channel found for id %d", id),
-	}
+	return nil, errors.NewNotFoundf("no channel found for id %d", id)
 }
 
 // CreateChannel creates an alert channel within a given account.
--- a/pkg/apm/labels.go
+++ b/pkg/apm/labels.go
@@ -1,6 +1,10 @@
 package apm
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/newrelic/newrelic-client-go/pkg/errors"
+)
 
 // ListLabels returns the labels within an account.
 func (apm *APM) ListLabels() ([]*Label, error) {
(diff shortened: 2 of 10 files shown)


pkg/alerts/channels.go
```diff
@@ -38,8 +40,9 @@ func (alerts *Alerts) GetChannel(id int) (*Channel, error) {
 			return channel, nil
 		}
 	}
-
-	return nil, fmt.Errorf("no channel found for id %d", id)
+	return nil, &errors.ErrorNotFound{
```
jthurman42: I know we've already implemented `errors.ErrorNotFound`, but it feels weird to read vs `errors.NotFound` / `errors.UnexpectedStatusCode`.

Before this PR, I only see a few references to them:

```
internal/http/client.go:		return nil, &errors.ErrorNotFound{}
internal/http/client_test.go:	assert.IsType(t, &errors.ErrorNotFound{}, err)

internal/http/client.go:		return nil, &errors.ErrorUnexpectedStatusCode{
internal/http/client_test.go:	assert.Equal(t, err.(*errors.ErrorUnexpectedStatusCode).Err, "error message")
internal/http/client_test.go:	assert.Equal(t, err.(*errors.ErrorUnexpectedStatusCode).Err, "error message")
internal/http/client_test.go:	assert.IsType(t, &errors.ErrorUnexpectedStatusCode{}, err)
```

pkg/apm/labels.go
```diff
@@ -39,7 +43,9 @@ func (apm *APM) GetLabel(key string) (*Label, error) {
 		}
 	}
 
-	return nil, fmt.Errorf("no label found with key %s", key)
+	return nil, &errors.ErrorNotFound{
+		Message: fmt.Sprintf("no label found with key %s", key),
```
jthurman42: This is happening throughout the PR, feels like a good place to make a `New` method, or something similar to `fmt.Errorf`

```
  return nil, errors.NewNotFoundf("no condition found for policy %d and condition ID %d", policyID, pluginID)
```

pkg/errors/errors.go
```diff
@@ -7,9 +7,15 @@ import (
 
 // ErrorNotFound is returned when a 404 response is returned
 // from New Relic's APIs.
-type ErrorNotFound struct{}
+type ErrorNotFound struct {
+	Message string
```
jthurman42: `ErrorUnexpectedStatusCode` uses `Err string` for this, not message.  Arguably should not be exported either, with a `New` and `Error()` functions the data within the struct can be hidden

jthurman42: Good reference for custom errors in Go: https://golangbot.com/custom-errors/

pkg/apm/labels.go
ctrombley: 👍

pkg/errors/errors.go
ctrombley: The terraform provider has [a need to check the error type](https://github.com/terraform-providers/terraform-provider-newrelic/pull/307/files#diff-39246a08dea1f12838e24b286d645fc8R219), so I think we'll need to export it.

pkg/alerts/channels.go
ctrombley: I updated the names of these types to omit the "Error" prefix.

pkg/errors/errors.go
```diff
@@ -5,22 +5,48 @@ import (
...
-	return fmt.Sprintf("404 not found")
+	return e.err
 }
 
-// ErrorUnexpectedStatusCode is returned when an unexpected
+// UnexpectedStatusCode is returned when an unexpected
 // status code is returned from New Relic's APIs.
-type ErrorUnexpectedStatusCode struct {
+type UnexpectedStatusCode struct {
 	Err        string
```
jthurman42: `err` like NotFound

ctrombley: 👍

Issue #67, opened 2020-01-06
Provide support for Entity Tags


ctrombley:
### Objective
Support the CRUD operations for Entity Tags as provided by Nerdgraph.

### Acceptance Criteria
- Support the entity tag mutations available via Nerdgraph:
  - `taggingAddTagsToEntity`
  - `taggingDeleteTagFromEntity`
  - `taggingDeleteTagValuesFromEntity`
  - `taggingReplaceTagsOnEntity`
- Establish a reusable pattern for implementing additional Nerdgraph functionality
- Take note of how this implementation might be generated via Nerdgraph schema introspection and gather notes and insights to that end.

sanderblue:
Resolved in #95

PR #95, opened 2020-01-15
feat(entities): add entities search and entity tagging

### LINKED ISSUE #66: Select a method of integrating with Nerdgraph
### Objective
Decide on a path forward for integrating with the features available in the Nerdgraph API.

### Acceptance Criteria
- Research the available options
  - https://github.com/shurcooL/graphql (Included on the graphql site under clients)
  - https://github.com/machinebox/graphql
  - net/http
- Select an approach to be used

ctrombley:
This PR introduces graphQL support and fills out entity and entity tag resource methods.

Resolves #66 and #67.

### DIFF
--- a/internal/http/graphql_client.go
+++ b/internal/http/graphql_client.go
@@ -1,0 +1,79 @@
+package http
+
+import (
+	"strings"
+
+	"github.com/newrelic/newrelic-client-go/internal/logging"
+	"github.com/newrelic/newrelic-client-go/internal/region"
+	"github.com/newrelic/newrelic-client-go/pkg/config"
+)
+
+// GraphQLClient represents a graphQL HTTP client.
+type GraphQLClient struct {
+	client *NewRelicClient
+	config config.Config
+	logger logging.Logger
+}
+
+// NewGraphQLClient returns a new instance of GraphQLClient.
+func NewGraphQLClient(cfg config.Config) *GraphQLClient {
+	if cfg.BaseURL == "" {
+		cfg.BaseURL = region.NerdGraphBaseURLs[region.Parse(cfg.Region)]
+	}
+
+	c := NewClient(cfg)
+	c.SetErrorValue(&graphQLErrorResponse{})
+
+	return &GraphQLClient{
+		client: &c,
+		config: cfg,
+		logger: cfg.GetLogger(),
+	}
+}
+
+// Query runs a graphQL query.
+func (g *GraphQLClient) Query(query string, vars map[string]interface{}, respBody interface{}) error {
+	req := graphQLRequest{
+		Query:     query,
+		Variables: vars,
+	}
+
+	resp := graphQLResponse{
+		Data: respBody,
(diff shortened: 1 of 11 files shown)


internal/http/graphql_client.go
```diff
@@ -0,0 +1,79 @@
...
+type GraphQLClient struct {
+	client *NewRelicClient
+	config config.Config
+	logger logging.Logger
+}
+
+// NewGraphQLClient returns a new instance of GraphQLClient.
+func NewGraphQLClient(cfg config.Config) *GraphQLClient {
+	if cfg.BaseURL == "" {
+		cfg.BaseURL = region.NerdGraphBaseURLs[region.Parse(cfg.Region)]
```
sanderblue: I realize this is an internal package and not publicly exposed (yet?), but should we try not to use proper nouns in the codebase?

ctrombley: Do you mean `NewGraphQLClient` vs `newGraphQLClient`?  The former is required to export the client so it can be used in the `entities` package.

sanderblue: I meant `NerdGraphBaseURLs` vs `GraphQLBaseURLs` 😎

internal/http/graphql_client.go
```diff
@@ -0,0 +1,79 @@
...
+	Query     string                 `json:"query"`
+	Variables map[string]interface{} `json:"variables,omitempty"`
+}
+
+type graphQLResponse struct {
+	Data interface{} `json:"data"`
+}
+
+type graphQLError struct {
+	Message string `json:"message"`
```
sanderblue: Thoughts on including the query `path` from the error? `Path []string `json:"path"``

i.e.
```
"message": "Entity guid required",
{
  //...
      "message": "Entity guid required"
      "path": [
        "actor",
        "entity"
      ]
}
```

ctrombley: Would you want to include it in the error message returned to the user?  Right now we are just doing a `strings.Join` on all the messages inside the returned errors, and returning that. 

https://github.com/newrelic/newrelic-client-go/pull/95/files#diff-f821ec257b1e058ba8dc4ba07b6edeceR75

sanderblue: Seems like that could be a noisy error response and potentially a little harder to read, but I haven't seen what the actual output looks like so hard to say. I think it would be a separate field in the current struct you have, but if it's a pain or involves anything more than a conditional check to see if `path` is present in the error object (not sure if its always present I guess), then it's not a big deal.

```go
type graphQLError struct {
	Message string `json:"message"`
  Path []string `json:"path"`
}
```

 I could see this as potentially helpful with debugging.

sanderblue: If it's unreasonable overhead to add to this struct
```go
type graphQLErrorResponse struct {
	Errors []graphQLError `json:"errors"`
}
```
then no worries.

Issue #101, opened 2020-01-18
Custom unmarshal method for Payload and Header fields in ChannelConfiguration struct


sanderblue:
### Objective
Placeholder for this [issue](https://github.com/terraform-providers/terraform-provider-newrelic/issues/323#issuecomment-575842788) in the Terraform provider

PR #102, opened 2020-01-21
fix(alerts): custom unmarshal of channel configuration Headers and Payload fields

sanderblue:
Resolves: #101 

### Synopsis
The New Relic v2 API endpoint for alert channels returns two different data types for `configuration.payload` and `configuration.headers`. If either of those fields is "empty" or left black when creating an alert channel in the UI, the API will return an empty string. But if headers and/or payload are included in an alert channel, the API will return an object, hence the need for a custom umarshal method for those fields.

**Example responses:**

Population headers and payload
```
{
  "type": "webhook",
  "configuration": {
    "headers": {                            
      "x-header-test": "header-value"  <-- notice we have an object here
    },                                                 
    "payload": {                                 
      "account_id": "$ACCOUNT_ID"   <-- notice we have an object here
    },
    "payload_type": "application/json",
    "base_url": "http://example.com"
  },
  // ...
}
```
Empty headers and payload
```
{
  "type": "webhook",
  "configuration": {
    "headers":  "",  <-- notice this is a string and NOT an object, so we have to guard for this
    "payload": "",   <-- notice this is a string and NOT an object, so we have to guard for this
    "payload_type": "application/json",
    "base_url": "http://example.com"
  },
  // ...
}
```

### DIFF
--- a/pkg/alerts/alerts_types.go
+++ b/pkg/alerts/alerts_types.go
@@ -1,6 +1,8 @@
 package alerts
 
 import (
+	"encoding/json"
+
 	"github.com/newrelic/newrelic-client-go/internal/serialization"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/alerts/alerts_types.go
```diff
@@ -18,25 +22,29 @@ type ChannelLinks struct {
...
+	RouteKey              string `json:"route_key,omitempty"`
+	ServiceKey            string `json:"service_key,omitempty"`
+	BaseURL               string `json:"base_url,omitempty"`
+	AuthUsername          string `json:"auth_username,omitempty"`
+	AuthPassword          string `json:"auth_password,omitempty"`
+	PayloadType           string `json:"payload_type,omitempty"`
+	Region                string `json:"region,omitempty"`
+	UserID                string `json:"user_id,omitempty"`
+
+	// Payload is unmarshaled to type map[string]string
```
ctrombley: "`Payload can be type converted to map[string]string`"?

sanderblue: Closing the loop. Discussed and we're good to go here 😎

PR #107, opened 2020-01-23
feat(newrelic): add the ability to configure base URLs per API

ctrombley:
This PR adds the ability to configure base URLs per API when using the rollup client in the `newrelic` package.

### DIFF
--- a/newrelic/newrelic.go
+++ b/newrelic/newrelic.go
@@ -139,7 +139,7 @@ func ConfigInfrastructureBaseURL(url string) ConfigOption {
 			return nil
 		}
 
-		return errors.New("Infrastructure base URL can not be empty")
+		return errors.New("infrastructure base URL can not be empty")
 	}
 }
 
@@ -151,7 +151,7 @@ func ConfigSyntheticsBaseURL(url string) ConfigOption {
 			return nil
 		}
 
-		return errors.New("Synthetics base URL can not be empty")
+		return errors.New("synthetics base URL can not be empty")
 	}
 }
 
@@ -163,7 +163,7 @@ func ConfigNerdGraphBaseURL(url string) ConfigOption {
 			return nil
 		}
 
-		return errors.New("NerdGraph base URL can not be empty")
+		return errors.New("nerdgraph base URL can not be empty")
 	}
 }
 
--- a/newrelic/newrelic.go
+++ b/newrelic/newrelic.go
@@ -118,7 +118,7 @@ func ConfigUserAgent(ua string) ConfigOption {
 	}
 }
 
-// ConfigBaseURL sets the Base URL used to make requests.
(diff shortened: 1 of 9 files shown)


newrelic/newrelic.go
```diff
@@ -129,3 +129,39 @@ func ConfigBaseURL(url string) ConfigOption {
...
+
+// ConfigInfrastructureBaseURL sets the base URL used to make requests to the Infrastructure API.
+func ConfigInfrastructureBaseURL(url string) ConfigOption {
+	return func(cfg *config.Config) error {
+		if url != "" {
+			cfg.InfrastructureBaseURL = url
+			return nil
+		}
+
+		return errors.New("Infrastructure base URL can not be empty")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

newrelic/newrelic.go
```diff
@@ -129,3 +129,39 @@ func ConfigBaseURL(url string) ConfigOption {
...
+
+// ConfigSyntheticsBaseURL sets the base URL used to make requests to the Synthetics API.
+func ConfigSyntheticsBaseURL(url string) ConfigOption {
+	return func(cfg *config.Config) error {
+		if url != "" {
+			cfg.SyntheticsBaseURL = url
+			return nil
+		}
+
+		return errors.New("Synthetics base URL can not be empty")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

newrelic/newrelic.go
```diff
@@ -129,3 +129,39 @@ func ConfigBaseURL(url string) ConfigOption {
...
+
+// ConfigNerdGraphBaseURL sets the base URL used to make requests to the NerdGraph API.
+func ConfigNerdGraphBaseURL(url string) ConfigOption {
+	return func(cfg *config.Config) error {
+		if url != "" {
+			cfg.NerdGraphBaseURL = url
+			return nil
+		}
+
+		return errors.New("NerdGraph base URL can not be empty")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

newrelic/newrelic.go
jthurman42: Looking at the provider, if these are not set by the user they would default to an empty string.  While it's a bit awkward, we should accept an empty string and ignore it.  This would be a good logging item to consider.

ctrombley: 👍

newrelic/newrelic.go
ctrombley: 👍

newrelic/newrelic.go
ctrombley: 👍

Issue #113, opened 2020-01-28
Alert Policy Channels: Cannot add more than one channel to a policy at a time


sanderblue:
### Description
Attempting to add multiple channels to an alert policy does not work. Only 

### Current behavior
When calling`alerts.UpdatePolicyChannels(321, []int{123345, 678899})` with multiple channel IDs only the last channel in the slice of IDs is added to the policy. The client generates a call that looks like the URL below.

Example API call:
```
PUT https://api.newrelic.com/v2/alerts_policy_channels.json?channel_ids=<id1>&channel_ids=<id2>&policy_id=<policyID>
```
Our current client implementation doesn't seem to handle the multiple instances of the query param `channel_ids`. Instead of create an array of IDs, it only uses the last one in the query string.

### Expected behavior
Attempting to add multiple channel IDs to an alert policy should ensure all IDs passed to the method are added to the associated policy ID.

### Steps To Reproduce
1. Call `alerts.UpdatePolicyChannels(policyID, []channelIDs)` with a known existing policy ID and _at least_ **2** known existing channel IDs. Can be set up in an integration test.
2. What the output in the terminal when running the test/call.
3. Check your New Relic account for search for the newly generated alert policy and notice that only one notification channel was added when there should be 2 channels added.

### References or Related Issues
This is needed to help support the ease of adding multiple channels to an alert policy via the Terraform provider.

PR #114, opened 2020-01-28
fix(alerts): ensure multiple channels can be added via alert policy channel endpoint

sanderblue:
Resolves: #113


sanderblue:
@ctrombley the library does handle it, but we need to use the `comma` option. I figured it had to handle it somehow, but I wasn't quite sure how/where to change the option. Good 'ol trial and error and we should be good now.

Issue #79, opened 2020-01-08
Provide support for Applications Metrics and Metrics Data endpoints


ctrombley:
The client should support the applications metrics and metrics data endpoionts: 
* https://rpm.newrelic.com/api/explore/applications/metric_names
* https://rpm.newrelic.com/api/explore/applications/metric_data

zlesnr:
I've got something started for this, and once I have push access I'll be able to put up a PR.

PR #115, opened 2020-01-28
feat(apm): Add support application metric names and data

zlesnr:
resolves: #79

Currently there is no facility in the code to query for the application metric
names or data.  Here we take a first pass implementation for the /metrics.json
and /metrics/data.json endpoints of the APM API.

### DIFF
--- a/pkg/apm/applications.go
+++ b/pkg/apm/applications.go
@@ -2,16 +2,8 @@ package apm
 
 import (
 	"fmt"
+	"time"
 )
-
-// ListApplicationsParams represents a set of filters to be
-// used when querying New Relic applications.
-type ListApplicationsParams struct {
-	Name     string `url:"filter[name],omitempty"`
-	Host     string `url:"filter[host],omitempty"`
-	IDs      []int  `url:"filter[ids],omitempty,comma"`
-	Language string `url:"filter[language],omitempty"`
-}
 
 // ListApplications is used to retrieve New Relic applications.
 func (apm *APM) ListApplications(params *ListApplicationsParams) ([]*Application, error) {
@@ -49,13 +41,6 @@ func (apm *APM) GetApplication(applicationID int) (*Application, error) {
 	return &response.Application, nil
 }
 
-// UpdateApplicationParams represents a set of parameters to be
-// used when updating New Relic applications.
-type UpdateApplicationParams struct {
-	Name     string
-	Settings ApplicationSettings
-}
-
 // UpdateApplication is used to update a New Relic application's name and/or settings.
 func (apm *APM) UpdateApplication(applicationID int, params UpdateApplicationParams) (*Application, error) {
 	response := applicationResponse{}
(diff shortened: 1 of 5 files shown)


pkg/apm/applications.go
```diff
@@ -88,6 +89,85 @@ func (apm *APM) DeleteApplication(applicationID int) (*Application, error) {
...
+		metrics = append(metrics, response.Metrics...)
+
+		paging := apm.pager.Parse(resp)
+		nextURL = paging.Next
+	}
+
+	return metrics, nil
+}
+
+// MetricData is used to retreive a list of values for each of the requested metrics.
```
golangcibot: `retreive` is a misspelling of `retrieve` (from `misspell`)
```suggestion

```

pkg/apm/applications.go
```diff
@@ -88,6 +73,54 @@ func (apm *APM) DeleteApplication(applicationID int) (*Application, error) {
 	return &response.Application, nil
 }
 
+// MetricNames is used to retrieve a list of known metrics and their value names for the given resource.
+//
+// https://rpm.newrelic.com/api/explore/applications/metric_names
```
sanderblue: Nice. We've only done this in a few places so far. I like the idea of a documentation pattern where we include links to docs if a method maps directly to an endpoint/doc like this. 😎

pkg/apm/applications_integration_test.go
```diff
@@ -3,114 +3,44 @@
 package apm
 
 import (
-	"os"
 	"testing"
 
-	"github.com/newrelic/newrelic-client-go/pkg/config"
+	"github.com/stretchr/testify/assert"
```
sanderblue: Not a blocker, but we noticed that using `github.com/stretchr/testify/require` seems slightly better suited for integration tests. It has the same API as `assert`, but in situations where the assertion fails, `require` will stop execution and short circuit the test, which ensures the next step of the integration test won't happen (bc it would likely cause another error). Helps limit the noise with error output when running integration tests.

pkg/apm/applications_test.go
```diff
@@ -190,3 +274,88 @@ func TestDeleteApplication(t *testing.T) {
...
+			assert.Equal(t, expected[i], actual[i])
+		}
+	}
+}
+
+func TestMetricData(t *testing.T) {
+	t.Parallel()
+	apm := newMockResponse(t, testMetricDataJson, http.StatusOK)
+
+	type Values struct {
```
sanderblue: If we added this struct to the `apm_types.go` it could be used for the `MetricTimeslice.Values` field as well as here. Thoughts?

pkg/apm/applications.go
```diff
@@ -88,6 +73,54 @@ func (apm *APM) DeleteApplication(applicationID int) (*Application, error) {
...
+func (apm *APM) MetricNames(applicationID int, params MetricNamesParams) ([]*MetricName, error) {
```
sanderblue: Can we name this `GetMetricNames()`? We've been prefixing the "action type" to our client methods - i.e. Create*, List*, Get*, Update*, Delete*

pkg/apm/applications.go
zlesnr: Let's hope the link for the docs doesn't change :)

pkg/apm/applications_test.go
zlesnr: That sounds good.  I think I had that in my head when I was writing it and then forgot.

pkg/apm/applications.go
zlesnr: Sounds reasonable to me.  I can do the same for the `MetricData`.

pkg/apm/applications_integration_test.go
zlesnr: Easy change.

zlesnr:
I believe I've addressed all the comments in 6cd4d5e.  Let me know if you notice anything else @sanderblue.

Issue #72, opened 2020-01-07
feat: Multi-location Synthetics Conditions


jthurman42:
### Objective
Add support for the recently GA Multi-location Synthetics Condition type.  Documentation: https://rpm.newrelic.com/api/explore/alerts_location_failure_conditions/create

### Acceptance Criteria
- New Condition type created
- Ability to CRUD the condition type

PR #118, opened 2020-01-30
feat(alerts): Implement multi-location synthetics conditions

zlesnr:
resolves: #72

Currently there is no facility to manage location_failure_conditions in the
Alerts API.  Here we add what is necessary to manage these resources.

### DIFF
--- a/pkg/alerts/alerts_types.go
+++ b/pkg/alerts/alerts_types.go
@@ -192,6 +192,29 @@ type SyntheticsCondition struct {
 	MonitorID  string `json:"monitor_id,omitempty"`
 }
 
+// MultiLocationSyntheticsCondition represents a location-based failure condition.
+//
+// ViolationTimeLimitSeconds must be one of 3600, 7200, 14400, 28800, 43200, 86400.
+type MultiLocationSyntheticsCondition struct {
+	ID                        int                                    `json:"id,omitempty"`
+	Name                      string                                 `json:"name,omitempty"`
+	Enabled                   bool                                   `json:"enabled"`
+	RunbookURL                string                                 `json:"runbook_url,omitempty"`
+	MonitorID                 string                                 `json:"monitor_id,omitempty"`
+	Entities                  []string                               `json:"entities,omitempty"`
+	Terms                     []MultiLocationSyntheticsConditionTerm `json:"terms,omitempty"`
+	ViolationTimeLimitSeconds int                                    `json:"violation_time_limit_seconds,omitempty"`
+}
+
+// MultiLocationSyntheticsConditionTerm represents a single term for a location-based failure condition.
+//
+// Priority must be "warning" or "critical".
+// Threshold must be greater than zero.
+type MultiLocationSyntheticsConditionTerm struct {
+	Priority  string `json:"priority,omitempty"`
+	Threshold int    `json:"threshold,omitempty"`
+}
+
 // MapStringInterface is used for custom unmarshaling of
 // fields that have potentially dynamic types.
 // E.g. when a field can be a string or an object/map
--- a/pkg/alerts/multi_location_synthetics_conditions.go
+++ b/pkg/alerts/multi_location_synthetics_conditions.go
@@ -1,0 +1,94 @@
+package alerts
+
+import (
+	"fmt"
+)
+
+// ListMultiLocationSyntheticsConditions returns alert conditions for a specified policy.
+func (alerts *Alerts) ListMultiLocationSyntheticsConditions(policyID int) ([]*MultiLocationSyntheticsCondition, error) {
+	response := multiLocationSyntheticsConditionListResponse{}
+	multiLocationSyntheticsConditions := []*MultiLocationSyntheticsCondition{}
+	queryParams := listMultiLocationSyntheticsConditionsParams{
+		PolicyID: policyID,
+	}
+
+	nextURL := fmt.Sprintf("/alerts_location_failure_conditions/policies/%d.json", policyID)
+
+	for nextURL != "" {
+		resp, err := alerts.client.Get(nextURL, &queryParams, &response)
+
+		if err != nil {
+			return nil, err
+		}
+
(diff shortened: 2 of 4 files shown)


pkg/alerts/location_failure_conditions.go
```diff
@@ -0,0 +1,96 @@
...
+	resp := locationFailureConditionCreateResponse{}
+
+	u := fmt.Sprintf("/alerts_location_failure_conditions/%d.json", condition.ID)
+	_, err := alerts.client.Put(u, nil, &reqBody, &resp)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// resp.Condition.PolicyID = condition.PolicyID
```
ctrombley: @zlesnr I think we've gone back and forth on this a couple times.  I don't think I see a problem with including the policy ID in the condition definition, though I'm pretty sure we didn't do it for the other condition types.  If you want to add it, go ahead but let's enter an issue to make it consistent in other condition types as well.

pkg/alerts/alerts_types.go
```diff
@@ -190,6 +190,24 @@ type SyntheticsCondition struct {
...
+	Name                      string                         `json:"name,omitempty"`
+	Enabled                   bool                           `json:"enabled"`
+	RunbookURL                string                         `json:"runbook_url,omitempty"`
+	MonitorID                 string                         `json:"monitor_id,omitempty"`
+	Entities                  []string                       `json:"entities,omitempty"`
+	Terms                     []LocationFailureConditionTerm `json:"terms,omitempty"`
+	ViolationTimeLimitSeconds int                            `json:"violation_time_limit_seconds,omitempty"`
+}
+
+// LocationFailureConditionTerm represents a single term for a location-based failure condition.
```
ctrombley: Should we include a note here about how to configure these?  In terms of how `Threshold` counts location failures?

pkg/alerts/location_failure_conditions.go
```diff
@@ -0,0 +1,96 @@
+package alerts
+
+import (
+	"fmt"
+)
+
+// ListLocationFailureConditions returns alert conditions for a specified policy.
```
ctrombley: For some of these resources we have been providing a `Get` (single) command as well that is just based on the `List` command.  Not a blocker but we might consider doing that here as well.

zlesnr:
Thanks for the review @ctrombley.  I agree the current names are a bit long, but I sort of picked that out of what the API docs call.  For the API Explorer, I too had trouble finding the docs there, but it seems that if you select an API key at the top first, then it shows additional documentation.  Seems like an odd UI thing.  I'm open to whatever we decide for naming the methods.

pkg/alerts/location_failure_conditions.go
zlesnr: I mistakenly left that comment in there from the resource that I copied.  Commented here because the object itself doesn't have a PolicyID, and I didn't go digging to see if the source object I copied did or not.  Definitely consistency, though here I think modeling the API(s) make sense.  I can dig a little more on what the other resources are doing there.

pkg/alerts/alerts_types.go
zlesnr: Yes, good point.  The docs in the API explorer are not quite complete, since what is meant by `integer` in some cases really means, the number of seconds in intervals of 1, 2, 4, 8 hours.  I have a slack message with some more accurate docs that I could transpose here.  I was wondering about how the official docs get updated, but in lieu of proper docs, at least notes on the method would be good.

pkg/alerts/location_failure_conditions.go
zlesnr: For this particular API I didn't see a get method, based on the API Explorer.  I can check with the source team, but I'd left it out since the API didn't look to support it.

pkg/alerts/location_failure_conditions.go
zlesnr: It looks like the PolicyID on an `alerts.Condition` also has a PolicyID attached to the object that is not returned from the API.

https://rpm.newrelic.com/api/explore/alerts_conditions/list

I'm slightly of the opinion that we shouldn't add extra fields to objects, but maybe there is a good use case to keep the policy ID attached to the condition.  I'll file a follow up issue to get consistency.

pkg/alerts/alerts_types.go
zlesnr: I've added a bit of information in the couple of places that were confusing like this one.  I'd been wondering about validating the input before we make the call, or if we instead just rely on good error messaging back from the API to return to the caller.  Maintaining that validation would put us in the middle of updating the code each time there were a change to the API, which doesn't sound great, but would provide us the opportunity to give the caller perhaps a fast and perhaps more informative error in the case where the API error returned is less than ideal.

pkg/alerts/location_failure_conditions.go
zlesnr: https://github.com/newrelic/newrelic-client-go/issues/120 to follow up.

pkg/alerts/location_failure_conditions.go
zlesnr: I've confirmed with the team that there is not a Get api for this, but there will be in the upcoming graphql api.

ctrombley:
@zlesnr I spoke with @jthurman42, let's replace instances of "LocationFailureCondition" with "MultiLocationSyntheticsCondition".

pkg/alerts/alerts_types.go
```diff
@@ -192,6 +192,29 @@ type SyntheticsCondition struct {
 	MonitorID  string `json:"monitor_id,omitempty"`
 }
 
+// MultiLocationSyntheticsCondition represents a location-based failure condition.
+//
+// ViolationTimeLimitSeconds must be one of 3600, 7200, 14400, 28800, 43200, 86400.
+type MultiLocationSyntheticsCondition struct {
+	ID                        int                            `json:"id,omitempty"`
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

zlesnr:
@ctrombley I've made the rename as requested.  Let me know if anything else stands out here.

PR #126, opened 2020-02-04
fix: allow string representations of JSON for alert channel webhook and payload

ctrombley:
This PR fixes unmarshaling of alert channel webhook headers and payload when the API returns string representations of JSON objects, like so:

```json
"configuration": {
   "headers": "{\"key\":\"value\"}",
   "payload": "{\"key\":\"value\"}",
},
```

### DIFF
--- a/pkg/alerts/alerts_types.go
+++ b/pkg/alerts/alerts_types.go
@@ -2,6 +2,8 @@ package alerts
 
 import (
 	"encoding/json"
+	"strconv"
+	"strings"
 
 	"github.com/newrelic/newrelic-client-go/internal/serialization"
 )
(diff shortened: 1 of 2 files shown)


pkg/alerts/alerts_types.go
```diff
@@ -201,11 +203,23 @@ type mapStringInterfaceProxy MapStringInterface
...
+
 	// Check for empty JSON string
-	if string(data) == `""` {
+	if s == `""` {
 		return nil
 	}
 
+	// Remove quotes if this is a string representation of JSON
+	if strings.HasPrefix(s, "\"") && strings.HasSuffix(s, "\"") {
+		s, err := strconv.Unquote(s)
```
golangcibot: shadow: declaration of "s" shadows declaration at line 206 (from `govet`)

ctrombley: 👍

PR #132, opened 2020-02-18
feat(ci): the beginnings of some release automation

sanderblue:
This PR introduces a new `make release` target to facilitate the release process. The [runbook](https://source.datanerd.us/developer-toolkit/runbooks/wiki/New-Relic-Go-SDK) has also been updated.

Example `make` command:
- `make release version=0.11.0`

### DIFF
--- a/build/release.mk
+++ b/build/release.mk
@@ -1,6 +1,6 @@
-RELEASE_SCRIPT_PATH ?= ./tools/release.sh
+RELEASE_SCRIPT ?= ./tools/release.sh
 
 # Example usage: make release version=0.11.0
 release:
 	@echo "=== $(PROJECT_NAME) === [ release          ]: Generating release."
-	./tools/release.sh $(version)
+	$(RELEASE_SCRIPT) $(version)
--- a/build/release.mk
+++ b/build/release.mk
@@ -1,0 +1,6 @@
+RELEASE_SCRIPT_PATH ?= ./tools/release.sh
+
+# Example usage: make release version=0.11.0
+release:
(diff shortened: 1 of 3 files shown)


build/release.mk
```diff
@@ -0,0 +1,6 @@
+RELEASE_SCRIPT_PATH ?= ./tools/release.sh
+
+# Example usage: make release version=0.11.0
```
zlesnr: Is there a way that we can require `version` as an argument?  This is me not knowing make well enough.

build/release.mk
```diff
@@ -0,0 +1,6 @@
...
+release:
+	@echo "=== $(PROJECT_NAME) === [ release          ]: Generating release."
+	./tools/release.sh $(version)
```
zlesnr: Maybe use `RELEASE_SCRIPT_PATH` from above in place of the direct path.  What do you think?

build/release.mk
sanderblue: It is required. If it's not passed in, an error message is displayed to the user.

build/release.mk
sanderblue: Oh yeah, I meant to do that (jet lag fever and such 😉). Thanks.

build/release.mk
zlesnr: Right, of course.

build/release.mk
sanderblue: Updated

zlesnr:
Nice work @sanderblue.  Thanks for putting this up.

PR #146, opened 2020-02-26
feat(http): allow personal API keys to be used for alerts and APM resources

ctrombley:
This PR allows Personal API keys to be used for authentication in the V2 endpoints that allow it (currently only the resources defined in the `apm` and `alerts` packages).

Compatibility mode will be enabled if a personal API key has been provided and the mode has been enabled within the package instantiating the client.

### DIFF
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -32,8 +32,15 @@ var (
 
 // NewRelicClient represents a client for communicating with the New Relic APIs.
 type NewRelicClient struct {
-	Client     *retryablehttp.Client
-	Config     config.Config
+	// Client represents the underlying HTTP client.
+	Client *retryablehttp.Client
+
+	// Config is the HTTP client configuration.
+	Config config.Config
+
+	// UsePersonalAPIKeyCompatability is for internal use only.
+	UsePersonalAPIKeyCompatibility bool
+
 	errorValue ErrorResponse
 }
 
@@ -202,6 +209,10 @@ func (c *NewRelicClient) setHeaders(req *retryablehttp.Request) {
 
 	if c.Config.PersonalAPIKey != "" {
 		req.Header.Set("Api-Key", c.Config.PersonalAPIKey)
+
+		if c.UsePersonalAPIKeyCompatibility {
+			req.Header.Set("Auth-Type", "User-Api-Key")
+		}
 	}
 }
 
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -39,7 +39,7 @@ type NewRelicClient struct {
 	Config config.Config
 
 	// UsePersonalAPIKeyCompatability is for internal use only.
-	UsePersonalAPIKeyCompatibility bool
+	AuthStrategy RequestAuthorizer
 
 	errorValue ErrorResponse
 }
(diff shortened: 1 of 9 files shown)


pkg/alerts/alerts_test.go
```diff
@@ -45,15 +45,21 @@ func newMockResponse(
 // nolint
 func newIntegrationTestClient(t *testing.T) Alerts {
 	apiKey := os.Getenv("NEWRELIC_API_KEY")
+	personalAPIKey := os.Getenv("NEWRELIC_PERSONAL_API_KEY")
 
-	if apiKey == "" {
+	if apiKey == "" && personalAPIKey == "" {
 		t.Skipf("acceptance testing requires an API key")
```
jthurman42: `requires NEWRELIC_API_KEY and NEWRELIC_PERSONAL_API_KEY`

internal/http/client.go
```diff
@@ -32,8 +32,15 @@ var (
...
-	Client     *retryablehttp.Client
-	Config     config.Config
+	// Client represents the underlying HTTP client.
+	Client *retryablehttp.Client
+
+	// Config is the HTTP client configuration.
+	Config config.Config
+
+	// UsePersonalAPIKeyCompatability is for internal use only.
+	UsePersonalAPIKeyCompatibility bool
```
jthurman42: Not sure that compatibility is the right item here, could be just `UsePersonalAPIKey`

internal/http/client.go
```diff
@@ -202,6 +209,10 @@ func (c *NewRelicClient) setHeaders(req *retryablehttp.Request) {
```
jthurman42: Should not send the APIKey header if explicitly said to use the Personal API Key

pkg/apm/apm.go
```diff
@@ -18,10 +18,11 @@ type APM struct {
 func New(config config.Config) APM {
 	pkg := APM{
 		client: http.NewClient(config),
-		logger: config.GetLogger(),
```
jthurman42: was this intentional?

pkg/apm/apm_test.go
```diff
@@ -42,13 +42,19 @@ func newMockResponse(
 // nolint
 func newIntegrationTestClient(t *testing.T) APM {
 	apiKey := os.Getenv("NEWRELIC_API_KEY")
+	personalAPIKey := os.Getenv("NEWRELIC_PERSONAL_API_KEY")
 
-	if apiKey == "" {
+	if apiKey == "" && personalAPIKey == "" {
 		t.Skipf("acceptance testing requires an API key")
```
jthurman42: Same as test above (improved error message)

pkg/apm/apm.go
```diff
@@ -18,10 +18,11 @@ type APM struct {
...
 		pager:  &http.LinkHeaderPager{},
 	}
 
+	pkg.client.UsePersonalAPIKeyCompatibility = true
```
zlesnr: I like this pattern for feature flags.

internal/http/client.go
ctrombley: `PreferPersonalAPIKey`?  It won't be used if it's not passed

pkg/apm/apm.go
ctrombley: Nope!  good catch

internal/http/client.go
ctrombley: 👍

PR #147, opened 2020-02-27
feat(nerdgraph): implement ability to make raw GraphQL query

sanderblue:
Resolves: #149 

### POC for raw GraphQL via client

First pass at adding the ability to pass raw GraphQL queries directly via a client method. 

This facilitates being able to execute a CLI command that passes a raw GraphQL query to the client.

#### CLI command example (basic query, no variables)
```
newrelic graphql query "{ actor { user { email id name } } }"
```

#### CLI command example (complex, with variables)
```
newrelic graphql query 'query($guid: EntityGuid!) { actor { entity(guid: $guid) { guid name domain entityType } } }' --variables '{"guid": "MjUyMDUyOHxBUE18QVBQTElDQVRJT058MjE1MDM3Nzk1"}'
```

### DIFF
--- a/newrelic/newrelic.go
+++ b/newrelic/newrelic.go
@@ -11,6 +11,7 @@ import (
 	"github.com/newrelic/newrelic-client-go/pkg/config"
 	"github.com/newrelic/newrelic-client-go/pkg/dashboards"
 	"github.com/newrelic/newrelic-client-go/pkg/entities"
+	"github.com/newrelic/newrelic-client-go/pkg/nerdgraph"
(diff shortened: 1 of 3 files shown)


newrelic/newrelic.go
```diff
@@ -23,6 +24,7 @@ type NewRelic struct {
 	Entities   entities.Entities
 	Plugins    plugins.Plugins
 	Synthetics synthetics.Synthetics
+	GraphQL    graphql.GraphQL
```
sanderblue: Possibly rename this package to `nerdgraph` given the current pattern is package by New Relic product.

PR #151, opened 2020-02-28
Fix json tags for ConditionTerm struct

davidji99:
Resolves https://github.com/davidji99/newrelic-client-go/issues/1 <-- Issue and context here.

### DIFF
--- a/pkg/alerts/alerts_types.go
+++ b/pkg/alerts/alerts_types.go
@@ -75,10 +75,10 @@ type ConditionUserDefined struct {
 
 // ConditionTerm represents the terms of a New Relic alert condition.
 type ConditionTerm struct {
-	Duration     int     `json:"duration,string,omitempty"`
+	Duration     int     `json:"duration,omitempty"`
 	Operator     string  `json:"operator,omitempty"`
 	Priority     string  `json:"priority,omitempty"`
-	Threshold    float64 `json:"threshold,string"`
+	Threshold    float64 `json:"threshold,omitempty"`
 	TimeFunction string  `json:"time_function,omitempty"`
 }
 


davidji99:
@sanderblue Pinging you since I saw you also work on the newrelic provider.

zlesnr:
Hi, thank you for the PR.  I don't think this is quite right.  Using the API explorer I see the response from the API looks like the following:

```
{
  "conditions": [
    {
      "id": 11468515,
      "type": "apm_app_metric",
      "name": "foot",
      "enabled": true,
      "entities": [
        "237778141"
      ],
      "metric": "apdex",
      "runbook_url": "https://www.example.com",
      "condition_scope": "application",
      "terms": [
        {
          "duration": "5",
          "operator": "below",
          "priority": "critical",
          "threshold": "0.75",
          "time_function": "all"
        }
      ]
    }
  ]
}
```

The Duration and the Threshold are strings in the response, but we deserialize them into the struct as Float and Integer.  The "string" tags in the struct assist with this data transformation.  The change proposed here would mean that we wouldn't be able to read the response back from the API.  In case you are interested, I believe this should be available to you: https://rpm.newrelic.com/api/explore/alerts_conditions/list

For the terraform issue you are seeing, would you mind opening an issue with an example manifest?

sanderblue:
@zlesnr thanks for following up here. 

I also confirmed the API response contains strings, hence the need to for some additional serialization help as mentioned.

I made a couple attempts to reproduce the problem, but was unable to get reproduce the error.

+1 on opening an issue in the New Relic Terraform Provider [issues](https://github.com/terraform-providers/terraform-provider-newrelic/issues). Please include your Terraform config (sensitive data redacted), your version of the provider, and any additional information that could be helpful. 🙂

davidji99:
@zlesnr Thanks for the information! I didn't realize this is method for unmarshalling `string` float values in Go.

@sanderblue I'll open an issue in the provider's repo. Thanks.

PR #155, opened 2020-02-28
BREAKING CHANGE: Change API Key field names

jthurman42:
Align API key names with our public docs.

This breaks:
* `ConfigAPIKey` as it becomes `ConfigAdminAPIKey`
* `NEWRELIC_PERSONAL_API_KEY` becomes `NEW_RELIC_API_KEY` <= Personal API Key
* `NEWRELIC_API_KEY` becomes `NEW_RELIC_ADMIN_API_KEY` <= Admin API Key

### DIFF
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -38,7 +38,7 @@ type NewRelicClient struct {
 	// Config is the HTTP client configuration.
 	Config config.Config
 
-	// UsePersonalAPIKeyCompatability is for internal use only.
+	// AuthStrategy allows us to use multiple authentication methods for API calls
 	AuthStrategy RequestAuthorizer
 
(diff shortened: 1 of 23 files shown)


jthurman42:
Updated CircleCI to use the new ENV vars, and re-ran the tests.

internal/http/client.go
```diff
@@ -38,7 +38,7 @@ type NewRelicClient struct {
 	// Config is the HTTP client configuration.
 	Config config.Config
 
-	// UsePersonalAPIKeyCompatability is for internal use only.
+	// AuthStrategy allows us to use multiple authentication methods for API calls
```
ctrombley: Thanks

PR #159, opened 2020-03-02
feat(entities): provide ability to specify return fields for entitySearch GraphQL query results

sanderblue:
This PR facilitates specifying which entity data to return for the entity search results. Sometimes a consumer might only care about a specific entity property such as `guid`. This will allow the consumer to specific "only return the `guid` property" for each entity in the search results versus returning all the properties for each entity.

### DIFF
--- a/pkg/entities/entity.go
+++ b/pkg/entities/entity.go
@@ -6,17 +6,17 @@ import (
 
 // Entity represents a New Relic One entity.
 type Entity struct {
-	AccountID  int
-	Domain     EntityDomainType
-	EntityType EntityType
-	GUID       string
-	Name       string
-	Permalink  string
-	Reporting  bool
-	Type       string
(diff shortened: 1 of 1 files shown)


pkg/entities/entity.go
```diff
@@ -146,62 +152,86 @@ func (e *Entities) GetEntity(guid string) (*Entity, error) {
...
+	fields := strings.Join(f[:], "\n")
+
+	return fmt.Sprintf(`
+		query($queryBuilder: EntitySearchQueryBuilder, $cursor: String) {
+			actor {
+				entitySearch(queryBuilder: $queryBuilder) {
+					results(cursor: $cursor) {
+						nextCursor
+						entities {
+							%s
```
jthurman42: This type of filtering won't work for things that are on an interface, like `ApplicationID` for APM is (see `searchEntitiesQuery`

Issue #120, opened 2020-01-31
Gain consistency about setting PolicyID on objects


zlesnr:
### Objective
Currently, some objects around `alerts.Condition` and soon to be `alerts.MultiLocationSyntheticsCondition` are inconsistent about how they make use of the PolicyID.  In both cases, the API object does not contain the PolicyID, but in the case of `alerts.Condition`, we create an extra field for `PolicyID` and attache the ID of the policy to the object after the call is made, before returning the object to the user.

We should determine which approach we want to take here and make the code consistent.

### Acceptance Criteria
- PolicyID fields that are not present in the API should be managed on conditions, or not.
- Usage consistency of PolicyID fields.

jthurman42:
Other alert condition types (all that I can tell) have PolicyID either in them or stitched in, so it seems reasonable to do so here as well.

sanderblue:
I _think_ one use case and reason why `PolicyID` was stitched in to some struct was to facilitate `terraform import` for alert conditions, NRQL alert conditions, and plugins conditions. 

Given the following HCL, it would be required to also have the alert policy defined as a separate block to import, but maybe someone didn't want to have both in their HCL.

```hcl
resource "newrelic_alert_condition" "foo" {
  policy_id = 123456

  name            = "foo"
  type            = "apm_app_metric"
  entities        = [654321]
  metric          = "apdex"
  condition_scope = "application"

  term {
    duration      = 5
    operator      = "below"
    priority      = "critical"
    threshold     = 0.5
    time_function = "all"
  }
}
```
In order for `terraform import newrelic_alert_condition.foo ...` to work, since the API response for an alert channel doesn't contain a policy ID association anywhere, you would need to provide the policyID in some way.  And since the Terraform _read_ functionality doesn't have a way to read the `policy_id` directly from the HCL and `policy_id` does't exist in the state yet (hence the import), that's potentially one reason why it's stitched into the struct.  

The stitched in `PolicyID` field is also used for **create** and is included in the return struct which is then used to create the ID hash of the resource. 

If we removed `PolicyID` from the `alerts.Condition` struct, we would need to make sure we returned a tuple of the condition and the policy ID. 

**TL;DR** It would appear that stitching the `PolicyID` might be a convenience thing vs a necessary thing, but the convenience factor seems to be pretty nice.

sanderblue:
[...]

PR #172, opened 2020-03-06
refactor: remove transient PolicyID on conditions structs for 1:1 mapping to API responses

sanderblue:
Resolves: #120 

Downstream changes will need to be made in [terraform-providers/terraform-provider-newrelic](https://github.com/terraform-providers/terraform-provider-newrelic).

### DIFF
--- a/pkg/alerts/conditions_integration_test.go
+++ b/pkg/alerts/conditions_integration_test.go
@@ -53,7 +53,7 @@ func TestIntegrationConditions(t *testing.T) {
 
 	require.NoError(t, err)
 
-	testCondition.PolicyID = policy.ID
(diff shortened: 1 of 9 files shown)


pkg/alerts/conditions_integration_test.go
```diff
@@ -53,7 +53,7 @@ func TestIntegrationConditions(t *testing.T) {
 
 	require.NoError(t, err)
 
-	testCondition.PolicyID = policy.ID
+	// testCondition.PolicyID = policy.ID
```
ctrombley: Why comment rather than delete?

sanderblue: wups, just missed this one, sorry

PR #182, opened 2020-03-11
feat(workloads): add a workloads resource

### LINKED ISSUE #153: feat: NR1 workloads
Based on the `workload` nerdgraph mutation category and the `actor.account.workload` query path.

ctrombley:
Resolves #153 .

### DIFF
--- a/pkg/workloads/workload.go
+++ b/pkg/workloads/workload.go
@@ -1,0 +1,166 @@
+package workloads
+
+import (
+	"time"
+
+	"github.com/newrelic/newrelic-client-go/internal/serialization"
+	"github.com/newrelic/newrelic-client-go/pkg/errors"
+)
+
+// Workload represents a New Relic One workload.
+type Workload struct {
+	Account             *AccountReference            `json:"account,omitempty"`
+	CreatedAt           *serialization.EpochTime     `json:"created_at,omitempty"`
+	CreatedBy           *UserReference               `json:"created_by,omitempty"`
+	Entities            []*WorkloadEntityRef         `json:"entities,omitempty"`
+	EntitySearchQueries []*WorkloadEntitySearchQuery `json:"entitySearchQueries,omitempty"`
+	EntitySearchQuery   string                       `json:"entitySearchQuery,omitempty"`
+	GUID                *string                      `json:"guid,omitempty"`
+	ID                  *int                         `json:"id,omitempty"`
+	Name                *string                      `json:"name,omitempty"`
+	Permalink           *string                      `json:"permalink,omitempty"`
+	ScopeAccounts       *WorkloadScopeAccounts       `json:"scopeAccounts,omitempty"`
+	UpdatedAt           *serialization.EpochTime     `json:"updated_at,omitempty"`
+}
+
+// AccountReference represents the account this workload belongs to.
+type AccountReference struct {
+	ID   int    `json:"id,omitempty"`
+	Name string `json:"name,omitempty"`
+}
+
+// WorkloadEntityRef represents an entity referenced by this workload.
+type WorkloadEntityRef struct {
+	GUID string `json:"id,omitempty"`
+}
+
+// WorkloadEntitySearchQuery represents an entity search used by this workload.
+type WorkloadEntitySearchQuery struct {
+	CreatedAt *time.Time               `json:"createdAt,omitempty"`
+	CreatedBy *UserReference           `json:"createdBy,omitempty"`
+	ID        *int                     `json:"id,omitempty"`
+	Name      *string                  `json:"name,omitempty"`
+	Query     *string                  `json:"query,omitempty"`
+	UpdatedAt *serialization.EpochTime `json:"updatedAt,omitempty"`
+}
+
+// UserReference represents a user referenced by this workload's search query.
+type UserReference struct {
+	Email    string `json:"email,omitempty"`
+	Gravatar string `json:"gravatar,omitempty"`
+	ID       int    `json:"id,omitempty"`
+	Name     string `json:"name,omitempty"`
+}
+
+// WorkloadScopeAccounts represents the accounts used to scope this workload.
+type WorkloadScopeAccounts struct {
+	AccountIDs []*int `json:"accountIds,omitempty"`
+}
+
+// ListWorkloads retrieves a set of New Relic One workloads by their account ID.
+func (e *Workloads) ListWorkloads(accountID int) ([]*Workload, error) {
+	resp := listWorkloadsResponse{}
+	vars := map[string]interface{}{
+		"accountId": accountID,
+	}
+
+	if err := e.client.Query(listWorkloadsQuery, vars, &resp); err != nil {
+		return nil, err
+	}
+
+	if len(resp.Actor.Account.Workload.Collections) == 0 {
+		return nil, errors.NewNotFound("")
+	}
+
+	return resp.Actor.Account.Workload.Collections, nil
+}
+
+// GetWorkload retrieve a New Relic One workload by its ID.
+func (e *Workloads) GetWorkload(accountID int, workloadID int) (*Workload, error) {
+	resp := getWorkloadResponse{}
+	vars := map[string]interface{}{
+		"accountId": accountID,
+		"id":        workloadID,
+	}
+
+	if err := e.client.Query(getWorkloadQuery, vars, &resp); err != nil {
+		return nil, err
+	}
+
+	if resp.Actor.Account.Workload.Collection.ID == nil {
+		return nil, errors.NewNotFound("")
+	}
+
+	return &resp.Actor.Account.Workload.Collection, nil
+}
+
+const (
+	// graphqlWorkloadStructFields is the set of fields that we want returned on workload queries,
+	// and should map back directly to the Workload struct
+	graphqlEntityStructFields = `
+			account {
+				id
+				name
+			}
+			createdBy {
+				email
+				gravatar
+				id
+				name
+			}
+			createdAt
+			entities {
+				guid
+			}
+			entitySearchQueries {
+				createdAt
+				createdBy {
+					email
+					gravatar
+					id
+					name
+				}
+				name
+				id
+				query
+				updatedAt
+			}
+			entitySearchQuery
+			guid
+			id
+			permalink
+			name
+			scopeAccounts {
+				accountIds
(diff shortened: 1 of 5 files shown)


pkg/workloads/workload.go
```diff
@@ -0,0 +1,207 @@
...
+
+// ScopeAccounts represents the accounts used to scope this workload.
+type ScopeAccounts struct {
+	AccountIDs []*int `json:"accountIds,omitempty"`
+}
+
+// CreateInput represents the input parameters used for creating or updating a workload.
+type CreateInput struct {
+	// EntityGUIDs         []*string                 `json:"entityGuids,omitempty`
+	Entities            []*string                 `json:"entities,omitempty`
```
golangcibot: structtag: struct field tag `json:"entities,omitempty` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (from `govet`)

pkg/workloads/workload.go
```diff
@@ -0,0 +1,207 @@
...
+	EntitySearchQueries []*EntitySearchQueryInput `json:"entitySearchQueryInput,omitempty`
```
golangcibot: structtag: struct field tag `json:"entitySearchQueryInput,omitempty` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (from `govet`)

pkg/workloads/workload.go
```diff
@@ -0,0 +1,207 @@
...
+	GUID                *string                  `json:"guid,omitempty"`
+	ID                  *int                     `json:"id,omitempty"`
+	Name                *string                  `json:"name,omitempty"`
+	Permalink           *string                  `json:"permalink,omitempty"`
+	ScopeAccounts       *ScopeAccounts           `json:"scopeAccounts,omitempty"`
+	UpdatedAt           *serialization.EpochTime `json:"updated_at,omitempty"`
+}
+
+// AccountReference represents the account this workload belongs to.
+type AccountReference struct {
```
jthurman42: I think this is common on most entities, and I've thought a few times about changing over to it from `accountId` so we get the name.  Might be worth moving this to `entities.AccountReference` so we can leverage it elsewhere.

pkg/workloads/workload_integration_test.go
```diff
@@ -0,0 +1,77 @@
...
+	client := newIntegrationTestClient(t)
+
+	actual, err := client.CreateWorkload(testAccountID, &testCreateInput)
+
+	require.NoError(t, err)
+	require.NotNil(t, actual)
+}
+
+// nolint
+func newIntegrationTestClient(t *testing.T) Workloads {
```
jthurman42: How often do we use this? I know I've seen it before in this (maybe?) repo.  Might be worth moving to `internal/testing` package or what not (ala `newrelic-cli/internal/testcobra` package)

pkg/workloads/workloads.go
```diff
@@ -0,0 +1,25 @@
...
+
+// New returns a new client for interacting with New Relic One workloads.
+func New(config config.Config) Workloads {
+	return Workloads{
+		client: http.NewGraphQLClient(config),
+		logger: config.GetLogger(),
+	}
+}
+
+// BaseURLs represents the base API URLs for the different environments of the New Relic REST API V2.
```
jthurman42: Copy/pasta here (comment says REST v2, it's for NerdGraph)

pkg/workloads/workload.go
```diff
@@ -0,0 +1,207 @@
...
+	listWorkloadsQuery = `query($accountId: Int!) { actor { account(id: $accountId) { workload { collections {` +
+		graphqlWorkloadStructFields +
+		` } } } } }`
+
+	createWorkloadMutation = `
+		mutation($accountId: Int!, $workload: WorkloadCreateInput!) {
+			workloadCreate(accountId: $accountId, workload: $workload) {` +
+		graphqlWorkloadStructFields +
+		` } }`
+)
```
jthurman42: 👍

pkg/workloads/workload.go
jthurman42: Missing `"` on these two (also remove the commented out EntityGUIDs)

pkg/workloads/workload.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// EntityRef represents an entity referenced by this workload.
+type EntityRef struct {
+	GUID string `json:"id,omitempty"`
+}
+
+// EntitySearchQuery represents an entity search used by this workload.
+type EntitySearchQuery struct {
+	CreatedAt *time.Time               `json:"createdAt,omitempty"`
```
jthurman42: `*serialization.EpochTime` ?

pkg/workloads/workload.go
```diff
@@ -0,0 +1,307 @@
...
+			workloadDelete(accountId: $accountId, workload: $workload) {` +
+		graphqlWorkloadStructFields +
+		` } }`
+
+	duplicateWorkloadMutation = `
+		mutation($accountId: Int!, $workload: WorkloadDuplicateInput!) {
+			workloadDuplicate(accountId: $accountId, workload: $workload) {` +
+		graphqlWorkloadStructFields +
+		` } }`
+	j
```
golangcibot: `j` is unused (from `deadcode`)

pkg/workloads/workload.go
```diff
@@ -0,0 +1,307 @@
...
+// DuplicateInput represents the input object used to identify the workload to be duplicated.
+type DuplicateInput struct {
+	Name       string  `json:"name,omitempty"`
+	SourceGUID *string `json:"sourceGuid,omitempty"`
+}
+
+// UpdateInput represents the input object used to identify the workload to be updated and its new changes.
+type UpdateInput struct {
+	// EntityGUIDs         []*string                 `json:"entityGuids,omitempty`
+	Entities            []*string                 `json:"entities,omitempty`
```
golangcibot: structtag: struct field tag `json:"entities,omitempty` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (from `govet`)

pkg/workloads/workload.go
```diff
@@ -0,0 +1,307 @@
...
+	EntitySearchQueries []*EntitySearchQueryInput `json:"entitySearchQueryInput,omitempty`
```
golangcibot: structtag: struct field tag `json:"entitySearchQueryInput,omitempty` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (from `govet`)

pkg/workloads/workload.go
ctrombley: The commented out field is the new one after Entities gets deprecated, so I wanted to leave it in.

pkg/workloads/workload.go
ctrombley: 👍

pkg/workloads/workload_integration_test.go
ctrombley: It's specific to the `workloads` package, though.  I like the idea of keeping all that code together.

pkg/workloads/workload.go
ctrombley: That was my first attempt, but it turns out it's not returned as an epoch.

pkg/workloads/workload_integration_test.go
jthurman42: Dah, right, I completely missed the return value...

pkg/workloads/workload.go
jthurman42: But UpdatedAt is?  We should raise that with the Workloads group if true

pkg/workloads/workload.go
jthurman42: 👍

pkg/workloads/workload.go
jthurman42: Entity has both:
```
{
  actor {
    entity(guid: "MjUwODI1OXxBUE18QVBQTElDQVRJT058MjA0MjYxNDEw") {
      account {
        id
        name
      }
      accountId
    }
  }
}
```
As well as the few interfaces I looked at.

pkg/workloads/workload.go
```diff
@@ -0,0 +1,307 @@
...
+		` } } } } }`
+
+	createWorkloadMutation = `
+		mutation($accountId: Int!, $workload: WorkloadCreateInput!) {
+			workloadCreate(accountId: $accountId, workload: $workload) {` +
+		graphqlWorkloadStructFields +
+		` } }`
+
+	deleteWorkloadMutation = `
+		mutation($accountId: Int!, $workload: WorkloadDeleteInput!) {
```
zlesnr: Where are the graphql types like `WorkloadDeleteInput` defined?  I was writing queries today and didn't understand this.  I'm trying to do something similar for another endpoint.

jthurman42: `WorkloadDeleteInput!` above is the GraphQL type, not Golang.  Look up at L156, where the query is called `var` is passed in as a `map[string]interface{}` where the key is the name of the required variable ("accountId" and "workload") for this example.

`workload` is of Go type `*DeleteInput` defined on L81 which matches Go style better (full name workloads.DeleteInput)

zlesnr: Thanks for the explanation.  The variable passing I think I understand, and I guess the types available are just known from a query like this one?
```
{
  __schema {
    types {
      name
    }
  }
}
```

ctrombley: I was trying to avoid the redundancy of types that share the package name like `workloads.WorkloadDeleteInput`.  Once we start generating this code I wonder if it would be better to just forget about it.

pkg/workloads/workload.go
ctrombley: I did and they are looking into it.

pkg/workloads/workload.go
jthurman42: @zlesnr Yes, or building the query via https://api.newrelic.com/graphiql which provides context-sensitive help including type definitions.

@ctrombley I agree with the non-redundant version, and I believe one of the linters looks for these as it's considered sub-optimal.

pkg/workloads/workload.go
jthurman42: 👍

pkg/workloads/workload.go
ctrombley: Moved to the `nerdgraph` package.

Issue #184, opened 2020-03-12
feat: NerdGraph API: Alerts policies


jthurman42:
### Feature Description
We currently create alert policies via the REST v2 API.  AI Ops recently released the ability to manage them through NerdGraph instead.  

### Additional context
https://docs.newrelic.com/docs/alerts/alerts-nerdgraph/nerdgraph-examples/nerdgraph-api-alerts-policies

PR #186, opened 2020-03-13
feat(alerts): implement graphql policy methods

zlesnr:
resolves: #184

### DIFF
--- a/pkg/alerts/alerts_test.go
+++ b/pkg/alerts/alerts_test.go
@@ -15,14 +15,12 @@ import (
 func newTestClient(handler http.Handler) Alerts {
 	ts := httptest.NewServer(handler)
 
-	c := New(config.Config{
+	return New(config.Config{
 		AdminAPIKey:           "abc123",
 		BaseURL:               ts.URL,
 		InfrastructureBaseURL: ts.URL,
 		UserAgent:             "newrelic/newrelic-client-go",
 	})
-
-	return c
 }
 
 // nolint
@@ -50,11 +48,9 @@ func newIntegrationTestClient(t *testing.T) Alerts {
 		t.Skipf("acceptance testing requires NEW_RELIC_API_KEY and NEW_RELIC_ADMIN_API_KEY")
 	}
 
-	client := New(config.Config{
+	return New(config.Config{
 		AdminAPIKey:    adminAPIKey,
 		PersonalAPIKey: personalAPIKey,
 		LogLevel:       "debug",
 	})
-
-	return client
 }
--- a/pkg/alerts/policies.go
+++ b/pkg/alerts/policies.go
@@ -33,20 +33,42 @@ type Policy struct {
 	UpdatedAt          *serialization.EpochTime `json:"updated_at,omitempty"`
 }
 
-// ListPoliciesParams represents a set of filters to be
-// used when querying New Relic alert policies.
+// QueryPolicy is similar to a Policy, but the resulting NerdGraph objects are
+// string IDs in the JSON response.
+type QueryPolicy struct {
+	ID                 int                    `json:"id,string"`
+	IncidentPreference IncidentPreferenceType `json:"incidentPreference"`
+	Name               string                 `json:"name"`
+}
+
+type QueryPolicyInput struct {
+	IncidentPreference IncidentPreferenceType `json:"incidentPreference"`
+	Name               string                 `json:"name"`
+}
+
+type QueryPolicyCreateInput struct {
+	QueryPolicyInput
+}
+
+type QueryPolicyUpdateInput struct {
+	QueryPolicyInput
+}
+
+// ListPoliciesParams represents a set of filters to be used when querying New
+// Relic alert policies.
 type ListPoliciesParams struct {
 	Name string `url:"filter[name],omitempty"`
 }
 
 // ListPolicies returns a list of Alert Policies for a given account.
-func (alerts *Alerts) ListPolicies(params *ListPoliciesParams) ([]Policy, error) {
+func (a *Alerts) ListPolicies(params *ListPoliciesParams) ([]Policy, error) {
 	alertPolicies := []Policy{}
+
 	nextURL := "/alerts_policies.json"
 
 	for nextURL != "" {
 		response := alertPoliciesResponse{}
-		resp, err := alerts.client.Get(nextURL, &params, &response)
+		resp, err := a.client.Get(nextURL, &params, &response)
 
 		if err != nil {
 			return nil, err
@@ -54,7 +76,7 @@ func (alerts *Alerts) ListPolicies(params *ListPoliciesParams) ([]Policy, error)
 
 		alertPolicies = append(alertPolicies, response.Policies...)
 
-		paging := alerts.pager.Parse(resp)
+		paging := a.pager.Parse(resp)
 		nextURL = paging.Next
 	}
 
@@ -62,8 +84,8 @@ func (alerts *Alerts) ListPolicies(params *ListPoliciesParams) ([]Policy, error)
 }
 
 // GetPolicy returns a specific alert policy by ID for a given account.
-func (alerts *Alerts) GetPolicy(id int) (*Policy, error) {
-	policies, err := alerts.ListPolicies(nil)
+func (a *Alerts) GetPolicy(id int) (*Policy, error) {
+	policies, err := a.ListPolicies(nil)
 
 	if err != nil {
 		return nil, err
@@ -79,13 +101,13 @@ func (alerts *Alerts) GetPolicy(id int) (*Policy, error) {
 }
 
 // CreatePolicy creates a new alert policy for a given account.
-func (alerts *Alerts) CreatePolicy(policy Policy) (*Policy, error) {
+func (a *Alerts) CreatePolicy(policy Policy) (*Policy, error) {
 	reqBody := alertPolicyRequestBody{
 		Policy: policy,
 	}
 	resp := alertPolicyResponse{}
 
-	_, err := alerts.client.Post("/alerts_policies.json", nil, &reqBody, &resp)
+	_, err := a.client.Post("/alerts_policies.json", nil, &reqBody, &resp)
 
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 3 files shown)


zlesnr:
The test failure here is addressed in #187.  I can rebase once merged.

pkg/alerts/alerts_test.go
```diff
@@ -50,11 +50,11 @@ func newIntegrationTestClient(t *testing.T) Alerts {
...
 
-	client := New(config.Config{
+	a := New(config.Config{
 		AdminAPIKey:    adminAPIKey,
 		PersonalAPIKey: personalAPIKey,
 		LogLevel:       "debug",
 	})
 
-	return client
+	return a
```
sanderblue: Not a blocker at all and not that it matters, but we could return early vs setting to a variable.

pkg/alerts/policies.go
```diff
@@ -33,37 +33,58 @@ type Policy struct {
...
+}
+
+// ListPoliciesParams represents a set of filters to be used when querying New
+// Relic alert policies.
 type ListPoliciesParams struct {
 	Name string `url:"filter[name],omitempty"`
 }
 
 // ListPolicies returns a list of Alert Policies for a given account.
-func (alerts *Alerts) ListPolicies(params *ListPoliciesParams) ([]Policy, error) {
```
jthurman42: Why the move away from the full name (`alerts` => `a`)?  We seem to be hit/miss on this, probably a good point to decide as a coding standard for our projects.

pkg/alerts/policies.go
```diff
@@ -113,19 +133,103 @@ func (alerts *Alerts) UpdatePolicy(policy Policy) (*Policy, error) {
...
+		vars := map[string]interface{}{
+			"accountID": accountID,
+			"policyIDs": []int{id},
+		}
+
+		if err := a.client.Query(alertPolicyQueryPolicy, vars, &resp); err != nil {
+			return nil, err
+		}
+
+		policies = append(policies, resp.Actor.Account.Alerts.PoliciesSearch.Policies...)
```
jthurman42: This is pretty deep reference without any error checking to make sure all of that exists.

pkg/alerts/policies.go
```diff
@@ -113,19 +133,103 @@ func (alerts *Alerts) UpdatePolicy(policy Policy) (*Policy, error) {
...
+
+		nextCursor = resp.Actor.Account.Alerts.PoliciesSearch.NextCursor
```
jthurman42: Same

pkg/alerts/policies_integration_test.go
```diff
@@ -7,13 +7,14 @@ import (
...
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestIntegrationPolicy(t *testing.T) {
+func TestAlertsPolicy_Legacy(t *testing.T) {
 	t.Parallel()
 
-	client := newIntegrationTestClient(t)
+	a := newIntegrationTestClient(t)
```
jthurman42: `a` isn't a very descriptive name compared to `client`

pkg/alerts/policies.go
```diff
@@ -113,19 +133,103 @@ func (alerts *Alerts) UpdatePolicy(policy Policy) (*Policy, error) {
...
+	resp := alertQueryPolicyCreateResponse{}
+
+	if err := a.client.Query(alertsPolicyCreatePolicy, vars, &resp); err != nil {
+		return nil, err
+	}
+
+	return &resp.QueryPolicy, nil
+}
+
+func (a *Alerts) QueryPolicyUpdate(accountID int, policyID int, policy QueryPolicyUpdateInput) (*QueryPolicy, error) {
```
sanderblue: Thoughts on naming mutation methods with `Mutation` prefix since it's a GraphQL mutation request? I realize the client only currently has a `client.Query()` method, which seems like it could be something more like `client.GraphQLRequest()` to be a little more semantic.

pkg/alerts/policies.go
```diff
@@ -113,19 +133,103 @@ func (alerts *Alerts) UpdatePolicy(policy Policy) (*Policy, error) {
...
+		if err := a.client.Query(alertPolicyQueryPolicy, vars, &resp); err != nil {
+			return nil, err
+		}
+
+		policies = append(policies, resp.Actor.Account.Alerts.PoliciesSearch.Policies...)
+
+		nextCursor = resp.Actor.Account.Alerts.PoliciesSearch.NextCursor
+	}
+
+	if len(policies) == 0 {
```
jthurman42: `switch len(policies)` instead?

pkg/alerts/policies.go
```diff
@@ -137,3 +241,64 @@ type alertPolicyResponse struct {
...
+	QueryPolicy QueryPolicy `json:"alertsPolicyCreate"`
+}
+
+type alertQueryPolicyUpdateResponse struct {
+	QueryPolicy QueryPolicy `json:"alertsPolicyCreate"`
+}
+type alertQueryPolicyResponse struct {
+	Actor struct {
+		Account struct {
+			Alerts struct {
```
jthurman42: Seems like `Alerts` should be the struct, then some GraphQL wrapper removal around it.  We should think of the best way to do this as it's an ongoing issue.  Ideally, we'll get to code-gen at some point and refactor all of these out.

pkg/alerts/policies.go
```diff
@@ -137,3 +241,64 @@ type alertPolicyResponse struct {
...
+		actor {
+			account(id: $accountID) {
+				alerts {
+					policiesSearch(searchCriteria: {
+						ids: $policyIDs
+					}) {
+						policies {
+							id
+							name
+							incidentPreference
```
jthurman42: Consider making this a `const` that is added to each of these other constants, so if suddenly there is a new field we have one place to update.

pkg/alerts/policies.go
```diff
@@ -137,3 +241,64 @@ type alertPolicyResponse struct {
...
+						} } } } } }`
+
+	alertsPolicyCreatePolicy = `mutation CreatePolicy($accountID: Int!, $policy: AlertsPolicyInput!){
```
jthurman42: Do we need the last `Policy` on these constants? Feels redundant :yodawg:

pkg/alerts/policies.go
jthurman42: `GraphQLRequest` seems reasonable (outside of this PR to change), Query was the direct port from the graphql client wrapper

pkg/alerts/policies.go
zlesnr: Short is my preference, single letter if possible that matches the object.  Also something that doesn't overlap with the package name also seems good.

sanderblue: Agreed. Seems like a pretty easy one to agree to a standardize for us. The alerts pkg seems to use `alerts *Alerts` and the entities pkg uses `e *Entities` 🤔 No strong preference to either one.

pkg/alerts/policies_integration_test.go
zlesnr: `client` isn't the client though, it's the Alerts object that has a `client` attached.

pkg/alerts/policies.go
zlesnr: I've replaced this with the query for just one resource to avoid the looping entirely.

pkg/alerts/policies.go
zlesnr: I see we do this for `data` currently, but is `Actor.Account` as prolific in our GQL responses?  I suppose we could case switch on the client, but then we're putting package level logic into the client.  Is there another approach you were thinking  of here?

pkg/alerts/policies.go
zlesnr: Not sure what you mean here.  Are you saying we do some string embedding for the constants?

pkg/alerts/policies.go
```diff
@@ -113,19 +133,81 @@ func (alerts *Alerts) UpdatePolicy(policy Policy) (*Policy, error) {
...
-
+	_, err := a.client.Delete(url, nil, &resp)
 	if err != nil {
 		return nil, err
 	}
 
 	return &resp.Policy, nil
 }
 
+func (a *Alerts) CreatePolicyMutatation(accountID int, policy QueryPolicyCreateInput) (*QueryPolicy, error) {
```
ctrombley: nit: typo here ("Mutatation")

pkg/alerts/policies.go
```diff
@@ -113,19 +133,81 @@ func (alerts *Alerts) UpdatePolicy(policy Policy) (*Policy, error) {
...
+	resp := alertQueryPolicyUpdateResponse{}
+
+	if err := a.client.Query(alertsPolicyUpdatePolicy, vars, &resp); err != nil {
+		return nil, err
+	}
+
+	return &resp.QueryPolicy, nil
+}
+
+// QueryPolicy queries graphql for a policy matching the given account ID and policy ID.
```
ctrombley: nit: "graphql" -> "NerdGraph"?

pkg/alerts/policies.go
```diff
@@ -113,19 +133,81 @@ func (alerts *Alerts) UpdatePolicy(policy Policy) (*Policy, error) {
...
+	}
+
+	if err := a.client.Query(alertPolicyQueryPolicy, vars, &resp); err != nil {
+		return nil, err
+	}
+
+	return &resp.Actor.Account.Alerts.Policy, nil
+}
+
+// DeletePolicyMutation is the graphql mutation to delete a policy given the account ID and the policy ID.
```
ctrombley: nit: "graphql" -> "NerdGraph" ?

pkg/alerts/policies.go
sanderblue: One quick way to check for consistency is in the [godocs](https://godoc.org/github.com/newrelic/newrelic-client-go/pkg/alerts)

pkg/alerts/policies.go
sanderblue: I think he means the variable name has `Policy` twice.

jthurman42: Yeah, `alertsPolicyCreatePolicy` could be `alertsPolicyCreate` (though we should also have something like @sanderblue said around gql / mutation / query.

pkg/alerts/policies.go
jthurman42: The `ANSI C` in me says `a` is totally fine, the rest of my dev senses say that single letter vars are bad :mkay:  This isn't a blocker, I know we do this elsewhere, and should start a standards doc...

pkg/alerts/policies.go
jthurman42: `actor` will always be there, it's the scope the request was made under to NerdGraph.  `Account` is an entire section of the NerdGraph schema.

This is fine for now, we need to look at a better way to generate structs from GQL schema so we aren't constantly building these out.

pkg/alerts/policies.go
```diff
@@ -137,3 +222,59 @@ type alertPolicyResponse struct {
 type alertPolicyRequestBody struct {
 	Policy Policy `json:"policy"`
 }
+
+type alertQueryPolicyCreateResponse struct {
```
jthurman42: `alert    Query    PolicyCreateResponse` . Query seems out of place with the rest of the structs starting with `alertsPolicy`.  `alertsPolicyCreateQueryResponse` reads more like the rest of our structs.

PR #187, opened 2020-03-13
refactor: Consolidate HTTP Clients

jthurman42:
Some refactoring that is in process:
* Renames the `internal/http.NewRelicClient` to `internal/http.Client`
* Consolidates `GraphQLClient` into `Client`, and exposes the `Query` function.
  This simplifies passing around clients for different endpoints, so now as we migrate we could call `Query` instead of `Get` or what not.
* Make all `http.Client` fields private, add a few helper methods

I think this is in a good direction, but not yet there.  Looking for some feedback.


internal/http/client.go
```diff
@@ -330,3 +290,35 @@ func makeRequestBodyReader(reqBody interface{}) (*bytes.Buffer, error) {
...
 	return b, nil
 }
+
+// GetBaseURL returns the BaseURL used by the client
+func (c *Client) GetBaseURL() string {
+	return c.config.BaseURL
+}
+
+// Query runs a graphQL query.
+func (c *Client) Query(query string, vars map[string]interface{}, respBody interface{}) error {
```
ctrombley: Does this belong in `graphql.go` as well?

jthurman42: Since it's a func on `Client` I would say no.

PR #188, opened 2020-03-15
docs(newrelic): add extensive godoc-based package-level docs and examples

ctrombley:
This PR introduces extended godoc-based package-level documentation and examples.


.golangci.yml
```diff
@@ -61,3 +61,5 @@ issues:
   max-per-linter: 0
   max-same-issues: 0
 
+skip-files:
+  - "example_.*.go"
```
jthurman42: Is this a required/recommended format (vs `*_example.go`)?

pkg/apm/example_keyTransaction_test.go
```diff
@@ -0,0 +1,40 @@
...
+		Name: "Example key transaction",
+	}
+
+	transactions, err := client.ListKeyTransactions(listParams)
+	if err != nil {
+		log.Fatal("error listing key transactions:", err)
+	}
+
+	// Get a key transaction by ID.
+	transaction, err := client.GetKeyTransaction(transactions[0].ID)
```
jthurman42: same magic index issue

pkg/dashboards/example_dashboard_test.go
```diff
@@ -0,0 +1,137 @@
...
+		Title: "Example dashboard",
+	}
+
+	dashboards, err := client.ListDashboards(listParams)
+	if err != nil {
+		log.Fatal("error listing dashboards:", err)
+	}
+
+	// Get dashboard by ID.
+	dashboard, err := client.GetDashboard(dashboards[0].ID)
```
jthurman42: magic index

pkg/entities/example_tags_test.go
```diff
@@ -0,0 +1,103 @@
...
+		},
+	}
+
+	entities, err := client.SearchEntities(searchParams)
+	if err != nil {
+		log.Fatal("error searching entities:", err)
+	}
+
+	// List the tags associated with a given entity.
+	entityGUID := entities[0].GUID
```
jthurman42: Magic index

pkg/apm/example_application_test.go
```diff
@@ -0,0 +1,58 @@
...
+		Name: "Example application",
+	}
+
+	apps, err := client.ListApplications(listParams)
+	if err != nil {
+		log.Fatal("error listing applications:", err)
+	}
+
+	// Get an application by ID.
+	app, err := client.GetApplication(apps[0].ID)
```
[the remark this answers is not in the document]
ctrombley: I figured it just has to compile.  It's for example only so I thought terse would be better than fully runtime checked.

pkg/apm/doc.go
```diff
@@ -0,0 +1,32 @@
...
+
+New Relic One entity tags are currently the preferred method for organizing your
+New Relic resources.  Consider using entity tags via the `entities` package if
+you are just getting started.  More information about entity tags and APM labels
+can be found at the following URL:
+
+https://docs.newrelic.com/docs/new-relic-one/use-new-relic-one/core-concepts/tagging-use-tags-organize-group-what-you-monitor#labels
+
+*/
+package apm
```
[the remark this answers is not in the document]
ctrombley: I like the doc.go setup, and I've seen it done in golang core.

.golangci.yml
ctrombley: This was unnecessary, I'll remove it.

pkg/apm/example_application_test.go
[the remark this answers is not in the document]
ctrombley: I'll add to the comment.

internal/http/client.go
```diff
@@ -61,11 +61,11 @@ func NewClient(cfg config.Config) Client {
 	}
 
 	if cfg.BaseURL == "" {
-		cfg.BaseURL = region.DefaultBaseURLs[region.Parse(cfg.Region)]
+		cfg.BaseURL = region.DefaultBaseURLs[region.Parse(string(cfg.Region))]
```
zlesnr: It would be neat if we didn't have to convert to strings here if region is a `Region` object.  We should be able to use them as keys in the map.  We can always come back to that later too, not trying to block.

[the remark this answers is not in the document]
ctrombley: `cfg.Region` isn't a region struct, it's a `RegionType` string-backed type. (An enum basically)

PR #197, opened 2020-03-18
feat(alerts): implement muting rules

zlesnr:


pkg/alerts/muting_rules.go
```diff
@@ -0,0 +1,175 @@
+package alerts
+
+// MutingRule represents the alert suppression mechanism in the Alerts API.
+type MutingRule struct {
+	ID          int                      `json:"id,string,omitempty"`
+	AccountID   int                      `json:"accountId,string,omitempty"`
```
sanderblue: Looks like `accountId ` is returned as type `Int` in the NerdGraph response. Do we want to unmarshal this to a string per the current JSON tag? Seems like we would want to unmarshal it to an `int`.

pkg/alerts/muting_rules.go
```diff
@@ -0,0 +1,175 @@
...
+
+	alertsMutingRulesCreate = `mutation CreateRule($accountID: Int!, $rule: AlertsMutingRuleInput!) {
+		alertsMutingRuleCreate(accountId: $accountID, rule: $rule) {
+			id
+		}
+	}`
+
+	alertsMutingRulesUpdate = `mutation UpdateRule($accountID: Int!, $ruleID: ID!, $rule: AlertsMutingRuleUpdateInput!) {
+		alertsMutingRuleUpdate(accountId: $accountID, id: $ruleID, rule: $rule) {
+			name
```
sanderblue: Thoughts on returning other fields here?

pkg/alerts/muting_rules.go
zlesnr: Good catch.  I'm surprised it worked at all.

pkg/alerts/muting_rules.go
zlesnr: Included description and enabled.

PR #199, opened 2020-03-19
fix(workloads): export the workloads API via the newrelic package

ctrombley:


pkg/workloads/workload.go
```diff
@@ -244,7 +244,7 @@ const (
 		` } }`
 
 	duplicateWorkloadMutation = `
-		mutation($accountId: Int!, $sourceGuid: EntityGuid!, $workload: WorkloadDuplicateInput!) {
+		mutation($accountId: Int!, $sourceGuid: EntityGuid!, $workload: WorkloadDuplicateInput) {
```
zlesnr: Does the `!` here just mean the field is required?

pkg/workloads/workloads.go
```diff
@@ -1,3 +1,5 @@
+// Package workloads provides a programmatic API for interacting with New Relic
```
zlesnr: Yum, package docs.

zlesnr: Oh I see there is also a `doc.go` above.  Are both points of documentation requried?

pkg/workloads/workload.go
sanderblue: Means "non-nullable", so basically yeah

pkg/workloads/workloads.go
ctrombley: No, just one will do... I'll clean this up in a future PR

PR #201, opened 2020-03-20
feat(nerdgraph): begin generating structs from schema

zlesnr:


pkg/nerdgraph/nerdgraph.go
```diff
@@ -41,8 +41,185 @@ func (n *NerdGraph) Query(query string, variables map[string]interface{}) (inter
...
+              }
+            }
+          }
+        }
+      }
+    }
+	`
+)
+
+// Wheee... :)
```
ctrombley: 😆

pkg/nerdgraph/typegen.go
```diff
@@ -0,0 +1,223 @@
...
+	// output collects each line of a struct type
+	output := []string{}
+
+	output = append(output, fmt.Sprintf("type %s struct {", t.Name))
+
+	// Fill in the struct fields for an input type
+	for _, f := range t.InputFields {
+		var fieldType string
+
+		fmt.Printf("handling kind %s: %+v\n\n", f.Type.Kind, f)
```
ctrombley: `log.Printf` for stderr?

zlesnr: Oh yeah, I wasn't sure the right way to get things to print due to being run as an integration test, though I think we probably want to move it out of a test and into a make target or something.  Just not sure where to locate the code that will do the command of generating the files.

PR #204, opened 2020-03-24
Move Region to pkg, nest all Regional URLs into it

jthurman42:
The end-game for this PR is to remove BaseURL logic from `internal/http`, and each package, and have them in a central place.  Motivation is to dumb-down the http client a bit so it's more reusable (instead of re-creating them frequently), and to expose the region data in a more consumable way.

Create a Region package containing all important info for all supported regions and their endpoints.

This exposes `pkg/region`, which contains the URL information for each backend API, per region.  You are still able to override specific URLs for specific endpoints as required.

For non REST / NerdGraph calls, the `*.client.Get|Put|Delete|Post|RawPast` calls have been changed to pass in the full URL for the region, and not rely on the internal HTTP client to do so.

This will require future refactoring to completely remove the host handling from the http client, and migrating all packages to use the full URL in all requests.


jthurman42:
Closed in favor of #221

PR #208, opened 2020-03-25
fix(nrql_conditions): fix spelling errors and types

sanderblue:
Adjustments that were needed and discovered while implementing NerdGraph functionality in the Terraform Provider.

**Related**: `terraform-provider-newrelic` [PR #487](https://github.com/terraform-providers/terraform-provider-newrelic/pull/487)

### DIFF
--- a/pkg/alerts/nrql_conditions.go
+++ b/pkg/alerts/nrql_conditions.go
@@ -6,14 +6,14 @@ import (
 	"github.com/newrelic/newrelic-client-go/pkg/errors"
 )
 
-// ThresholdOccurence specifies the threshold occurrence for NRQL alert condition terms.
-type ThresholdOccurence string
+// ThresholdOccurrence specifies the threshold occurrence for NRQL alert condition terms.
+type ThresholdOccurrence string
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/alerts/nrql_conditions.go
```diff
@@ -357,7 +357,7 @@ func (a *Alerts) CreateNrqlConditionBaselineMutation(
 // UpdateNrqlConditionBaselineMutation updates a baseline NRQL alert condition via New Relic's NerdGraph API.
 func (a *Alerts) UpdateNrqlConditionBaselineMutation(
 	accountID int,
-	conditionID string, // GraphQL scalar type `ID` is a string in JSON
+	conditionID int,
```
jthurman42, line 360: I don't think we should change this type.  Today IDs look like ints, but that could bite us down the road (which is why the scalar type ID in GraphQL is an opaque string)

sanderblue: Just closing the loop for reference later if needed. Spoke about this a bit more and we can keep it like this for since this is more consistent with what we're doing now. We shall see what "the future" holds 🔮

PR #210, opened 2020-03-25
feat(alerts): implement NerdGraph policy search

zlesnr:


pkg/alerts/policies.go
```diff
@@ -193,6 +197,32 @@ func (a *Alerts) QueryPolicy(accountID, id int) (*QueryPolicy, error) {
 	return &resp.Actor.Account.Alerts.Policy, nil
 }
 
+// QueryPolicySearch searches NerdGraph for policies.
+func (a *Alerts) QueryPolicySearch(accountID int, params AlertsPoliciesSearchCriteriaInput) ([]QueryPolicy, error) {
```
sanderblue: I think we've been returning an array of pointers - i.e. `[]*QueryPolicy` - for consistency

zlesnr: Done.

pkg/alerts/types.go
```diff
@@ -0,0 +1,161 @@
...
+	* **policyName** - The alert policy name
+	* **conditionName** - The alert condition name
+	* **conditionType** - The alert condition type, such as app_metric
+	* **conditionRunbookUrl** - The alert condition's runbook url
+	* **product** - The target product (e.g., SYNTHETICS)
+	* **targetId** - The ID of the alerts target
+	* **targetName** - The name of the alerts target
+	* **nrqlEventType** - The NRQL event type
+	* **tag** - Arbitrary tags associated with some entity (e.g., FACET from a NRQL query)
+	* **nrqlQuery** - The NRQL query string */
```
sanderblue: Do these show up in godocs?

jthurman42: The probably don't as the first word in the comment is not the Name of the object.  We should refactor the typegen to resolve that as a follow up

zlesnr: Let's find out.

PR #221, opened 2020-03-27
Update all packages to pass full URLs

jthurman42:
Replacement / continuation of #204 that implements passing base URLs to all REST packages.  Depends on the other PR being merged.


internal/http/request.go
```diff
@@ -36,34 +34,24 @@ func NewRequest(c Client, method string, url string, params interface{}, reqBody
 		authStrategy: c.authStrategy,
 	}
 
+	// FIXME: We should remove this requirement on the request
```
ctrombley: We should decide as a team on TODO or FIXME.. I prefer TODO as some IDEs/editor plugins have built-in support for TODO

internal/testing/config.go
```diff
@@ -0,0 +1,68 @@
...
+		cfg.Region().SetRestBaseURL(testServer.URL)
+		cfg.Region().SetSyntheticsBaseURL(testServer.URL)
+	}
+
+	return cfg
+}
+
+// NewIntegrationTestConfig grabs environment vars for required fields or skips the test.
+// returns a fully saturated configuration
+func NewIntegrationTestConfig(t *testing.T) config.Config {
```
ctrombley: 👍

newrelic/newrelic.go
```diff
@@ -85,10 +86,14 @@ func ConfigAdminAPIKey(adminAPIKey string) ConfigOption {
 }
 
 // ConfigRegion sets the New Relic Region this client will use.
-func ConfigRegion(region config.RegionType) ConfigOption {
+func ConfigRegion(r region.Name) ConfigOption {
 	return func(cfg *config.Config) error {
-		cfg.Region = region
-		return nil
+		if region, ok := region.Regions[r]; ok {
+			regCopy := *region
```
ctrombley: What's the need for the copy?  Just to safeguard against runtime changes to region?

pkg/alerts/example_policy_test.go
```diff
@@ -10,10 +10,9 @@ import (
 func Example_policy() {
 	// Initialize the client configuration.  A Personal API key or Admin API key
 	// is required to communicate with the backend API.
-	cfg := config.Config{
-		AdminAPIKey:    os.Getenv("NEW_RELIC_ADMIN_API_KEY"),
-		PersonalAPIKey: os.Getenv("NEW_RELIC_API_KEY"),
-	}
+	cfg := config.New()
+	cfg.AdminAPIKey = os.Getenv("NEW_RELIC_ADMIN_API_KEY")
+	cfg.PersonalAPIKey = os.Getenv("NEW_RELIC_API_KEY")
```
zlesnr: Is there a reason here not to have the `New()` call populate the API key fields?  If not, I might suggest this as a change, so that a a new user only need pay attention to `client := New(config.New())` to have a fully working config.

pkg/alerts/plugins_conditions.go
```diff
@@ -107,9 +107,9 @@ func (a *Alerts) UpdatePluginsCondition(condition PluginsCondition) (*PluginsCon
 // DeletePluginsCondition deletes a plugin alert condition.
 func (a *Alerts) DeletePluginsCondition(id int) (*PluginsCondition, error) {
 	resp := pluginConditionResponse{}
-	u := fmt.Sprintf("/alerts_plugins_conditions/%d.json", id)
+	url := fmt.Sprintf("/alerts_plugins_conditions/%d.json", id)
 
-	_, err := a.client.Delete(u, nil, &resp)
+	_, err := a.client.Delete(a.config.Region().RestURL(url), nil, &resp)
```
zlesnr: For these kinds of calls, I was seeing the parts passed to the `RestURL()` method as "path".  Was the intent here to have something like `a.config.Region().RestURL("alerts_plugins_conditions", string(id), "json")` ?

pkg/region/region_test.go
```diff
@@ -0,0 +1,92 @@
...
+		"us":      Regions[US],
+		"Us":      Regions[US],
+		"uS":      Regions[US],
+		"US":      Regions[US],
+		"eu":      Regions[EU],
+		"Eu":      Regions[EU],
+		"eU":      Regions[EU],
+		"EU":      Regions[EU],
+		"staging": Regions[Staging],
+		"Staging": Regions[Staging],
```
ctrombley: `STAGING`?

newrelic/newrelic.go
jthurman42: Exactly,  Since we can not make Struct constants, always copy because user can change the result and have side effects otherwise.

internal/http/request.go
jthurman42: Either way. `vim` highlights both 😄

pkg/alerts/plugins_conditions.go
jthurman42: Yes, to enable something like:
```
a.config.Region().RestURL("alerts_plugins_conditions", strconv.Itoa(id)+".json")
```

Note using `+` as all strings passed in will be joined with `/` otherwise.

pkg/alerts/example_policy_test.go
jthurman42: `New()` should not read from the Env, as what if you have a config you want and happen to have env vars set?  (Common use case when I'm doing dev work and switch to a demo real quick).

Having a `NewFromEnv` or something along those lines was a thought, though did not implement at this time.

pkg/alerts/plugins_conditions.go
zlesnr: Yeah, I like that.

PR #222, opened 2020-03-27
Reorganize typegen code in repo

jthurman42:
The current location of `cmd/typegen.go` causes issues with the build system (output of a binary `typegen.go`, which we then try to lint/etc because `*.go`).  To resolve, this PR follows guidance from https://github.com/golang-standards/project-layout#tools (our standard project layout guide).

* Creates a `tools` directory
* Moves `typegen.go` into the `tools` directory
* Moves `tools.go` into the `tools` directory
* Updates `make generate` with the new path
* Moves `typegen.yaml` into `configs/` to follow the standards doc


jthurman42:
@zlesnr I started reading through https://blog.golang.org/generate during a bout of insomnia, and the last commit reorganizes yet again to leverage `go generate`.

### To generate `pkg/*/types.go`
* Add typegen.yaml file to each package where generate should happen
* Add `//go:generate ../../bin/typegen -v` to the main package file

### Changes to do this
* Add building tool binaries to the tools make step
* Re-wrote `make generate` to use `go generate` to exec commands defined in the go files for all packages in the repo

### Future work
* typegen to accept package name via flag vs yaml (`$GOPACKAGE` is available)

jthurman42:
It might be worth looking at moving typegen completely out in the future to a generic utility that converts GraphQL schema to Go structs (since I haven't seen any) . That would prevent some of the `../../bin/typegen` stuff

PR #234, opened 2020-04-01
fix(region): log an error when a bad region is specified

zlesnr:


zlesnr:
Related to https://github.com/newrelic/newrelic-cli/issues/101

newrelic/newrelic.go
```diff
@@ -89,14 +91,15 @@ func ConfigAdminAPIKey(adminAPIKey string) ConfigOption {
...
 // ConfigRegion sets the New Relic Region this client will use.
-func ConfigRegion(r region.Name) ConfigOption {
+func ConfigRegion(r string) ConfigOption {
 	return func(cfg *config.Config) error {
-		if region, ok := region.Regions[r]; ok {
-			regCopy := *region
-			return cfg.SetRegion(&regCopy)
+
+		slimRegion := region.Parse(r)
+		if slimRegion.String() != r {
```
jthurman42, line 98: This feels awkward to have the user perform the check.  Instead, moving the logging logic into the `Parse` function would be more appropriate, and you could return an error though with log + default value I don't think it's needed.

zlesnr: Sure, I can move the message into the parse method.

PR #240, opened 2020-04-02
chore(nergraph): log the nerdgraph request body easier debug

zlesnr:


internal/http/client.go
```diff
@@ -308,6 +309,13 @@ func (c *Client) NewNerdGraphRequest(query string, vars map[string]interface{},
 		Data: respBody,
 	}
 
+	varB, _ := json.Marshal(vars)
+
+	q1 := strings.Replace(query, "\n", "", -1)
+	q2 := strings.Replace(q1, "\t", " ", -1)
+
+	c.config.GetLogger().Trace("NewNerdGraphRequest", "vars", string(varB), "query", q2)
```
jthurman42, line 317: Is this not similar to how we already log the body when the request is executed?

https://github.com/newrelic/newrelic-client-go/blob/master/internal/http/client.go#L214

zlesnr: It is, but by the time that log message shows up, the reqBody is nil because it's been converted to a io.Reader.  Here we're logging while we still have the data without having to copy the reader twice, or figure out what kind of interface it is and log it appropriately.

PR #243, opened 2020-04-02
feat(typegen): handle scalar types

zlesnr:


pkg/nerdgraph/typegen.go
```diff
@@ -25,7 +29,7 @@ func ResolveSchemaTypes(schema Schema, typeNames []string) (map[string]string, e
 }
 
 func handleEnumType(schema Schema, t SchemaType) map[string]string {
-	types := make(map[string]string)
+	typeKeeer := make(map[string]string)
```
ctrombley, line 32: typo?

pkg/nerdgraph/typegen.go
```diff
@@ -7,6 +7,10 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var (
+	types = make(map[string]string)
```
jthurman42, line 11: Does this need to be package global?

zlesnr: I'm trying to figure that out now.

pkg/nerdgraph/typegen.go
zlesnr: Yes.

zlesnr:
This is not currently generating the enums.  This should be fixed before merge.

pkg/nerdgraph/typegen.go
zlesnr: Currently, yes.  The recursion makes it such that the global is updated when a new type is seen, so we don't end up with duplicates, or multiple fetches for the same time.

zlesnr:
The code is now able to compile with the generated types.

PR #247, opened 2020-04-06
Split parsing / fetching of region data

jthurman42:
Previous behavior for `region.Parse(<string>)` was to return the entire region.  Since we have a use-case for validating only the String => region.Name

New behavior:
* `region.Default` is a Name (not the entire Region struct) and a constant
* `region.Parse(string)` returns the `region.Name` or an error
* `region.Get(Name)` returns a copy of the region from the map of valid regions (DRY of copy due to non-constant structs)

### DIFF
--- a/newrelic/newrelic.go
+++ b/newrelic/newrelic.go
@@ -5,8 +5,6 @@ import (
 	"net/http"
 	"time"
 
-	log "github.com/sirupsen/logrus"
-
 	"github.com/newrelic/newrelic-client-go/internal/logging"
 	"github.com/newrelic/newrelic-client-go/pkg/alerts"
 	"github.com/newrelic/newrelic-client-go/pkg/apm"
@@ -94,10 +92,7 @@ func ConfigAdminAPIKey(adminAPIKey string) ConfigOption {
 func ConfigRegion(r region.Name) ConfigOption {
 	return func(cfg *config.Config) error {
 		reg, err := region.Get(r)
-		if _, ok := err.(*region.UnknownUsingDefaultError); ok {
-			// Log
-			log.Error(err.Error())
-		} else {
+		if err != nil {
 			return err
 		}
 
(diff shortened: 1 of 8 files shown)


newrelic/newrelic.go
```diff
@@ -91,12 +93,17 @@ func ConfigAdminAPIKey(adminAPIKey string) ConfigOption {
 // ConfigRegion sets the New Relic Region this client will use.
 func ConfigRegion(r region.Name) ConfigOption {
 	return func(cfg *config.Config) error {
-		if region, ok := region.Regions[r]; ok {
-			regCopy := *region
-			return cfg.SetRegion(&regCopy)
+		reg, err := region.Get(r)
+		if _, ok := err.(*region.UnknownUsingDefaultError); ok {
+			// Log
+			log.Error(err.Error())
```
jthurman42, line 99: we do not log anywhere else, so not sure this feels right... Downstream could do this instead

zlesnr: This seems okay for now.  I suspect that we could return something of a composite error too if we wanted rather than logging here.  Like join  the two error strings with a `, ` or something.

jthurman42: Error wrapping is a thing, but I generally believe that the error returned could be passed all the way back and be understood by the end-user of the Client

zlesnr: Yeah, for sure agree about getting that error back to the caller.  I can't tell what it would break here, but we could set the default region and return the error about using the default region to the caller.  You could just have a `SetRegion()` call right above this log message, and instead of log, return the error.

jthurman42: I removed this.  We fail fast everywhere else, and I think that is the right thing to do for the Client.  We do hand back an error.  If we want to add resiliency to the client + logging later we can do so in more detail.

PR #252, opened 2020-04-13
Refactoring of typegen

jthurman42:
Late night sleepless work on refactoring `typegen`

### Why

Working with typegen to add some auto-generation for `nerdgraph` there were a few areas that I wanted to improve (specifically around overriding the default SCALAR types used in Go).  Then things kinda snowballed a bit from there...

### Changes

* Change config file format to include ability to override the Golang type used to create a scalar
* Add handling of interfaces as, `interface{}` so they are at least created though probably not what you want...
* Collapse SchemaField and SchemaInputField to a single type, then build functions off of it for generating the output.
* String concat (`+=`) instead of `[]string` where it makes sense

### TODO / Ideas
- [x] Better docs on how to use typegen
- [ ] Split fetch and generate into different code paths (would allow for caching the API call, pre-generating and sanitizing it)


tools/typegen/README.md
```diff
@@ -0,0 +1,72 @@
+# typegen
```
ctrombley, line 1: Great docs 👍

jthurman42:
Going to go about this differently.

PR #286, opened 2020-05-04
fix(alerts): Updating incorrect AlertEvents params

jhunthrop:
- Changing `ListAlertEventsParams` to the correct ones.

### DIFF
--- a/pkg/alerts/events.go
+++ b/pkg/alerts/events.go
@@ -1,8 +1,6 @@
 package alerts
 
 import (
-	"time"
-
 	"github.com/newrelic/newrelic-client-go/internal/serialization"
 )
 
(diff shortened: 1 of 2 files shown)


zlesnr:
Thanks for the PR.  Not sure about that failing security test, but I can look.  Since you were able to make the change to the object without changing a test, that says to me that there isn't a test to cover this object.  Is that something that you would be willing to contribute also?

jhunthrop:
Security tests are failing because of:
```
`snyk` requires an authenticated account. Please run `snyk auth` and try again.
```

Yes, can add tests for the filters.

zlesnr:
That failing snyk action looks to be a side effect of using github actions.  Just hadn't seen that since most of the contributions have been within NewRelic.

jhunthrop:
@zlesnr Open to a better way to test input parameters with mocked backend.

This bug was a result of my PR: #280 .

I was looking at the wrong call in the API explorer when modeling the `ListAlertEventsParams` struct.

zlesnr:
This is a good start.  Thank you for the contribution.

PR #304, opened 2020-05-15
feat(edge): add trace observer resource

ctrombley:
This PR adds New Relic Edge support.  

### Integration testing notes:
Only one trace observer can be created per master account and provider region, which blocks us from being able to use our integration testing account for this resource (since account 1 has already created a trace observer that I assume should not be deleted).  I introduced a `NEW_RELIC_MASTER_ACCOUNT_ID` environment variable to be used in the tests and marked the integration tests to be skipped if it is not set.   I then tested against my own personal master account, which also required the NEW_RELIC_API_KEY from that account.  Also note that access to edge tracing is still gated by a feature flag, which I had to enable for my account before testing successfully.

### Code generation notes:
Code gen was used for the data structures but manual changes were required afterward, so the generate directive was removed from the package after it was used.  Two issues required manual changes:

1. The ID fields assumed string data types, but integers were used in the NerdGraph schema (this was not addressed in this PR)
2. The same enum values appeared in different enums, causing naming collisions in the package namespace.  As a result, I updated the enum generation to use a slightly different structure that introduces a kind of namespacing.  An example of the changes can be seen in the [diff below](https://github.com/newrelic/newrelic-client-go/compare/feat/infinite-tracing?expand=1#diff-2a53df3ae9c90105f2feaea61ec03e7fL8-L19).


pkg/edge/trace_observers_integration_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	mock "github.com/newrelic/newrelic-client-go/internal/testing"
+)
+
+var (
+	testAccountIDStr      = os.Getenv("NEWRELIC_MASTER_ACCOUNT_ID")
```
jthurman42, line 14: Should be `NEW_RELIC_MASTER_ACCOUNT_ID` to go along with the NR standard env vars

tools/typegen/typegen.go
```diff
@@ -167,14 +167,18 @@ func (s *Schema) Definition(typeInfo TypeConfig) (string, error) {
...
 
+		output += "var " + t.Name + "Types = struct {\n"
 		for _, v := range t.EnumValues {
 			output += v.GetDescription()
-			output += "\t" + v.Name + " " + t.Name + " = \"" + v.Name + "\"\n"
+			output += "\t" + v.Name + " " + t.Name + "\n"
+		}
+		output += "}{\n"
+		for _, v := range t.EnumValues {
+			output += "\t" + v.Name + ": \"" + v.Name + "\",\n"
```
jthurman42, line 178: Bit of a change in how we do ENUMs (which I'm fine with, but want to call it out)

pkg/edge/trace_observers_integration_test.go
ctrombley: 👍

PR #322, opened 2020-05-22
HTTP Compression / Prep for Insights

jthurman42:
This PR accomplishes two goals:
* Add Insights URLs / building to Region
* Implement compression (none and gzip) in the http client, so sending data is optimized

The compression is not directly inline by default, however there is on spicy refactor: Removal of RawPost, and collapsing that logic into the [request](https://github.com/newrelic/newrelic-client-go/pull/322/files#diff-c8120cadd719e78b5424e31344a1b4e0R51).  In this case, anything that is already a `[]byte` is not marshaled to JSON, anything else is.  The one RawPost use-case has been updated (wrap: `[]byte("string:notJSON")`)

### DIFF
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -1,7 +1,6 @@
 package http
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -37,6 +36,9 @@ type Client struct {
 
 	// authStrategy allows us to use multiple authentication methods for API calls
 	authStrategy RequestAuthorizer
+
+	// compressor is used to compress the body of a request, and set the content-encoding header
+	compressor RequestCompressor
 
 	errorValue ErrorResponse
 }
@@ -81,6 +83,7 @@ func NewClient(cfg config.Config) Client {
 	return Client{
 		authStrategy: &ClassicV2Authorizer{},
 		client:       r,
+		compressor:   &NoneCompressor{},
 		config:       cfg,
 		errorValue:   &DefaultErrorResponse{},
 	}
@@ -90,6 +93,12 @@ func NewClient(cfg config.Config) Client {
 // which can be overridden per request
 func (c *Client) SetAuthStrategy(da RequestAuthorizer) {
 	c.authStrategy = da
+}
+
+// SetRequestCompressor is used to enable compression on the request using
+// the RequestCompressor specified
+func (c *Client) SetRequestCompressor(compressor RequestCompressor) {
+	c.compressor = compressor
 }
 
 // SetErrorValue is used to unmarshal error body responses in JSON format.
@@ -127,34 +136,6 @@ func (c *Client) Post(
 	respBody interface{},
 ) (*http.Response, error) {
 	req, err := c.NewRequest(http.MethodPost, url, queryParams, reqBody, respBody)
-	if err != nil {
-		return nil, err
-	}
-
-	return c.Do(req)
-}
-
-// RawPost behaves the same as Post, but without marshaling the body into JSON before making the request.
-// This is required at least in the case of Synthetics Labels, since the POST doesn't handle JSON.
-func (c *Client) RawPost(
-	url string,
-	queryParams interface{},
-	reqBody interface{},
-	respBody interface{},
-) (*http.Response, error) {
-
-	var requestBody []byte
-
-	switch val := reqBody.(type) {
-	case []byte:
-		requestBody = val
-	case string:
-		requestBody = []byte(val)
-	default:
-		return nil, errors.New("invalid request body")
-	}
-
-	req, err := c.NewRequest(http.MethodPost, url, queryParams, requestBody, respBody)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


internal/http/client.go
```diff
@@ -81,6 +83,7 @@ func NewClient(cfg config.Config) Client {
 	return Client{
 		authStrategy: &ClassicV2Authorizer{},
 		client:       r,
+		compressor:   &NoneCompressor{},
```
jthurman42, line 86: Someday this should be changed to `&GzipCompressor{}` by default.

internal/http/client.go
```diff
@@ -271,21 +252,6 @@ func isResponseSuccess(resp *http.Response) bool {
 	return statusCode >= http.StatusOK && statusCode <= 299
 }
 
-func makeRequestBodyReader(reqBody interface{}) (*bytes.Buffer, error) {
```
jthurman42, line 274: Collapsed this into the single calling function and refactored it.

internal/http/client.go
```diff
@@ -134,34 +143,6 @@ func (c *Client) Post(
 	return c.Do(req)
 }
 
-// RawPost behaves the same as Post, but without marshaling the body into JSON before making the request.
-// This is required at least in the case of Synthetics Labels, since the POST doesn't handle JSON.
-func (c *Client) RawPost(
```
jthurman42, line 139: Collapsed this logic into `Post`, where if `body` is of type `[]byte` it is not marshaled. The `string` case here can be done inline with `Post(...., []byte(strBody), ...)`

internal/http/compress.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"bufio"
+	"bytes"
+	"compress/gzip"
+	"io"
+)
+
+const (
+	// CompressorMinimumSize is the required size in bytes that a body exceeds before it is compressed
+	CompressorMinimumSize = 150
```
jthurman42, line 12: Played around with this to determine a reasonable size for JSON to be gzip'd. There is a test JSON is 153 bytes to validate compression does happen.

internal/http/compress_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	mock "github.com/newrelic/newrelic-client-go/pkg/testhelpers"
+)
+
+var testCompressionCases = []struct {
+	compress bool
```
jthurman42, line 16: If compress is false, we expect the data to NOT be compressed by the compressor, and instead compare data == data after running through the compressor.

internal/http/client.go
ctrombley: 👏

internal/http/compress_test.go
```diff
@@ -0,0 +1,96 @@
...
+	},
+	{
+		compress: false,
+		data:     []byte(`abcdefghijklmnopqrstuvwxyz1234567890`),
+	},
+	{
+		compress: true,
+		data:     []byte(`{"data": "json", "maybe":"didn't check", "handcrafted":true, "example": { "sub": "object", "arry": [ "why", "not" ] }, "todo": [ "make", "it", "larger" }`),
+		gzip: []byte{
+			0x1f, 0x8b, 0x8, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xff, 0x24, 0x8c, 0x4d, 0xae, 0xc2, 0x30, 0xc, 0x6, 0xaf, 0xf2,
```
ctrombley, line 32: 🤖

internal/http/request.go
```diff
@@ -67,6 +81,18 @@ func (r *Request) SetHeader(key string, value string) {
 	r.request.Header.Set(key, value)
 }
 
+// GetHeader returns the value of the header requested
+func (r *Request) GetHeader(key string) string {
+	return r.request.Header.Get(key)
+}
+
+// DelHeader returns the value of the header requested
```
ctrombley, line 89: s/returns/deletes/ ?

internal/http/request.go
```diff
@@ -67,6 +81,18 @@ func (r *Request) SetHeader(key string, value string) {
...
+func (r *Request) DelHeader(key string) {
```
ctrombley, line 90: How about `DeleteHeader`?  No strong opinion here, I like the symmetry with the underlying API also.

pkg/config/config.go
```diff
@@ -10,6 +10,21 @@ import (
 	"github.com/newrelic/newrelic-client-go/pkg/region"
 )
 
+// Compression to use during transport.
```
ctrombley, line 13: Comment should start with `CompressionType`?

pkg/region/region.go
```diff
@@ -130,7 +132,7 @@ func (r *Region) SetSyntheticsBaseURL(url string) {
 	}
 }
 
-// SyntheticsURL returns the Full URL for Infrastructure REST API Calls, with any additional path elements appended
+// SyntheticsURL returns the Full URL for Synthetics REST API Calls, with any additional path elements appended
```
ctrombley, line 135: 👍

internal/http/compress_test.go
jthurman42: copy pasta, must be getting hungry...

internal/http/request.go
jthurman42: Yeah, Del is what's under the hood which is why I left it.  Delete feels better, but so does `Header.Del` for phrasing.

PR #325, opened 2020-05-27
Use default User-Agent header if none provided

jthurman42:
This addresses the issue of incomplete user-agent headers by default:

* Calls to `config.New()` set the user-agent incorrectly (does not include full user-agent info like version)
* Add check to NewRequest to ensure that either the user-agent specified of the default is added to all requests.

### DIFF
--- a/internal/http/request.go
+++ b/internal/http/request.go
@@ -71,7 +71,12 @@ func (c *Client) NewRequest(method string, url string, params interface{}, reqBo
 
 	req.SetHeader(defaultNewRelicRequestingServiceHeader, cfg.ServiceName)
 	req.SetHeader("Content-Type", "application/json")
-	req.SetHeader("User-Agent", cfg.UserAgent)
+
+	if cfg.UserAgent != "" {
+		req.SetHeader("User-Agent", cfg.UserAgent)
+	} else {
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -59,7 +59,6 @@ func New() Config {
 
 	return Config{
 		region:      reg,
-		UserAgent:   "newrelic/newrelic-client-go",
```
jthurman42, line 62: Stop adding the incomplete header by default.

internal/http/request.go
```diff
@@ -71,7 +71,12 @@ func (c *Client) NewRequest(method string, url string, params interface{}, reqBo
 
 	req.SetHeader(defaultNewRelicRequestingServiceHeader, cfg.ServiceName)
 	req.SetHeader("Content-Type", "application/json")
-	req.SetHeader("User-Agent", cfg.UserAgent)
+
+	if cfg.UserAgent != "" {
+		req.SetHeader("User-Agent", cfg.UserAgent)
+	} else {
+		req.SetHeader("User-Agent", defaultUserAgent)
+	}
```
jthurman42, line 79: Ensure we have a header value to send, otherwise default.

PR #360, opened 2020-06-17
feat(eventstometrics): add EventsToMetrics

ctrombley:
Add support for events to metrics bridge.


pkg/eventstometrics/eventstometrics_integration_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	nr "github.com/newrelic/newrelic-client-go/pkg/testhelpers"
+)
+
+var (
```
sanderblue, line 13: Thoughts on moving this `var` declaration, scoping it to within the `TestIntegrationEventsToMetrics` test 
function?

pkg/eventstometrics/typegen.yaml
```diff
@@ -0,0 +1,14 @@
+---
+types:
```
sanderblue, line 2: 👍  nice

pkg/eventstometrics/eventstometrics_integration_test.go
ctrombley: 👍

PR #378, opened 2020-07-09
chore(build): integrate tutone as part of make generate

zlesnr:
relates to: https://github.com/newrelic/tutone/issues/14


.generate.yml
```diff
@@ -0,0 +1,17 @@
+---
+packages:
```
zlesnr, line 2: This config file will probably change shape before we want to merge here.

build/generate.mk
```diff
@@ -6,10 +6,16 @@ GO           ?= go
 
 PACKAGES ?= $(shell $(GO) list ./...)
 
+GOTOOLS += github.com/newrelic/tutone/cmd/tutone
+
 # Generate then lint fixes
-generate: generate-run lint-fix
+generate: generate-run generate-tutone lint-fix
+
+generate-tutone:
```
sanderblue, line 14: Not a blocker, but thoughts?

```suggestion
generate-types:
```

zlesnr: I'm thinking we will want to do more than types here pretty soon, so just referencing the tool here seems okay to me.

Issue #385, opened 2020-07-16
NRQL Alert Condition: setting threshold to 0 causes marshaling issues


sanderblue:
### Description
When creating a NRQL alert condition, setting a threshold value to `0` causes marshaling to omit the property instead of setting to `0.0` (it's a `float64` type) in JSON.

The `Threshold` field needs to be a pointer - i.e. `*float64` so the zero-value is marshaled properly.

### Current behavior
Currently the Terraform provider resource encounters an error when trying to provision a NRQL condition that has a term threshold of `0`.

### Expected behavior
The provider should successfully create the resource even if the threshold value is `0`.

### Additional Context
Oddly enough, we don't use `omitempty` in the JSON tag, so there could be something going on somewhere else, but a fix to this as a pointer should be okay.

PR #386, opened 2020-07-16
fix(nrql_conditions): better zero-value marshalling for threshold

sanderblue:
Resolves: #385

---
This PR also removes some extraneous test cases that are no longer relevant for monitor labels. Integration tests were failing for `TestIntegrationGetMonitorLabels` due to the following error. Tags are the future-facing resource. Labels are being removed and considered legacy.

> Error 403 Tag migration has already occurred for this account. Adding and removing labels is not allowed.


zlesnr:
This looks great! :goat:++

PR #403, opened 2020-07-24
chore: add test run reporting

ctrombley:


.github/workflows/test.yml
```diff
@@ -64,6 +64,13 @@ jobs:
     - name: Unit Tests
       run: make test-unit
 
+    - name: Post test results to New Relic
+      uses: newrelic/junit-reporter-action@v1
```
sanderblue, line 68: You might need to use `newrelic/junit-reporter-action@master` here since it's **v0.1.0**. Either that or make an official `v1.0.0` release of that GH action. Unfortunately I don't think the actions framework resolves minor versions the way we might expect.

ctrombley: Thanks!

PR #407, opened 2020-07-29
fix(region): make ConfigRegion case-insensitive

ctrombley:


newrelic/newrelic_test.go
```diff
@@ -61,6 +61,16 @@ func TestNew_setRegion(t *testing.T) {
...
 }
 
+func TestNew_setRegionCaseInsensitive(t *testing.T) {
+	t.Parallel()
+
+	nr, err := New(ConfigPersonalAPIKey(testAPIkey), ConfigRegion("staging"))
+
+	assert.NotNil(t, nr)
+	assert.NoError(t, err)
+	assert.Equal(t, "Staging", nr.config.Region().Name)
```
jthurman42, line 71: To keep `name` un-exported, this could be:

`assert.Equal(t, "Staging", nr.config.Region().String())`

pkg/region/region.go
```diff
@@ -18,7 +18,7 @@ type Name string
 
 // Region represents the members of the Region enumeration.
 type Region struct {
-	name                  string
+	Name                  string
```
jthurman42, line 21: Do we need to expose this? (vs using `String()` in the test case above). IIRC this was not exported to prevent the default set from being modified at run-time (since they are structs and can't be const)

newrelic/newrelic_test.go
ctrombley: 👍

pkg/region/region.go
ctrombley: 👍

newrelic/newrelic.go
```diff
@@ -115,9 +119,12 @@ func ConfigAdminAPIKey(adminAPIKey string) ConfigOption {
 }
 
 // ConfigRegion sets the New Relic Region this client will use.
-func ConfigRegion(r region.Name) ConfigOption {
+func ConfigRegion(r string) ConfigOption {
```
zlesnr, line 122: Could rename the method for clarity.  `ConfigureRegion()` perhaps?

ctrombley: I think we'd want to rename all of our config methods for consistency if we wanted to do this here.  How strongly do you feel about this?

ctrombley: Let's talk about adding this as a separate PR.

PR #426, opened 2020-08-13
chore(tutone): include action for code generation

zlesnr:


.github/workflows/generate.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Generate new code
+        run: make generate
+        env:
+          NEW_RELIC_API_KEY: ${{ secrets.NEW_RELIC_API_KEY }}
+
+      - name: Create Pull Request
+        id: cpr
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.PAT }}
```
zlesnr, line 18: This needs to use the correct token.

PR #427, opened 2020-08-14
Log api

msummers-nr:
Implement New Relic Log API


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+<a name="v0.36.0"></a>
+## [v0.36.0] - 2020-08-14
+### Features
+- **logs:** Implement Log API
```
sanderblue, line 4: Was this an automated change? Our CHANGELOG is updated via automation, so if this was manually added, we can remove it here.

sanderblue:
@msummers-nr This looks awesome 😎  After checking out your branch, integration tests passed on my end for the new feature. If you could either rebase or pull in the latest from the master branch and resolve the conflicts, we can make more moves on this today and hopefully get it merged.

pkg/logs/logs.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	rsp, err := l.client.Post(l.config.Region().LogsURL(), nil, logEntry, nil)
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(os.Stderr, "CreateLogEntry: rsp: %+v\n", rsp)
+	if !(rsp.StatusCode < 299) {
```
sanderblue, line 50: Does our internal client incorrectly handle an error status code? We have a check using `statusCode >= http.StatusOK && statusCode <= 299` to help determine whether or not the request was successful so just want to make sure this is behaving as expected.

pkg/region/region.go
```diff
@@ -187,3 +188,30 @@ func concatURLPaths(host string, path []string) (string, error) {
...
+
+// SetLogsBaseURL Allows overriding the Logs Base URL
+func (r *Region) SetLogsBaseURL(url string) {
+	if r != nil && url != "" {
+		r.logsBaseURL = url
+	}
+}
+
+// LogsURL returns the Full URL for the Log API (data ingest)
+func (r *Region) LogsURL(path ...string) string {
```
sanderblue, line 204: Does this need an argument? I see the function being called but without any args. Futuring-facing thoughts here maybe?

pkg/logs/logs.go
```diff
@@ -0,0 +1,55 @@
...
+		return errors.New("logs: CreateLogEntry: logEntry is nil, nothing to do")
+	}
+
+	rsp, err := l.client.Post(l.config.Region().LogsURL(), nil, logEntry, nil)
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(os.Stderr, "CreateLogEntry: rsp: %+v\n", rsp)
```
sanderblue, line 49: Not a blocker, but would we want to use our logger instead for this so this isn't always displayed in the output?

internal/http/auth.go
```diff
@@ -44,3 +44,9 @@ type InsightsInsertKeyAuthorizer struct{}
 func (a *InsightsInsertKeyAuthorizer) AuthorizeRequest(r *Request, c *config.Config) {
 	r.SetHeader("X-Insert-Key", c.InsightsInsertKey)
 }
+
+type LicenseKeyAuthorizer struct{}
+
+func (a *LicenseKeyAuthorizer) AuthorizeRequest(r *Request, c *config.Config) {
+	r.SetHeader("X-License-Key", c.LicenseKey)
+}
```
jthurman42, line 52: I'd be hesitant to add License key here.  IIRC Logs support sending via insert keys which already exist.

pkg/logs/logs.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// CreateLogEntry reports a log entry to New Relic.
+// It's up to the caller to send a valid Log API payload, no checking done here
+func (l *Logs) CreateLogEntry(logEntry interface{}) error {
+	if logEntry == nil {
+		return errors.New("logs: CreateLogEntry: logEntry is nil, nothing to do")
+	}
+
+	rsp, err := l.client.Post(l.config.Region().LogsURL(), nil, logEntry, nil)
```
jthurman42, line 43: Is the intent to block on every call to send a log?  Batching log lines together, queuing, and background sending while a bit more complex definitely would scale better.  The Event package implements this format.

internal/http/auth.go
msummers-nr: Their preferred method is License Key and I had no luck getting an Insert Key to work

pkg/logs/logs.go
msummers-nr: That's a nice to have an beyond the scope of the project.

internal/http/auth.go
jthurman42: 😢

pkg/logs/logs.go
msummers-nr: No it doesn't. Plus the simple pass/fail that it attempts isn't all that great- more detail at the level of the caller would be a good thing. Wrapping http status seems like a wrong-minded abstraction.

pkg/region/region.go
msummers-nr: We typically have more than one endpoint per region, for Logs there's currently one per. So yes.

pkg/logs/logs.go
msummers-nr: My mistake. Needs to be fixed.

internal/http/auth.go
msummers-nr: I agree, seems like leaving sharp knives out.

msummers-nr:
Closing as I'm unable to fix the broken commit message.

PR #429, opened 2020-08-17
Implement new endpoints for managing license keys

davidji99:
Resolves https://github.com/davidji99/newrelic-client-go/issues/2

This is a WIP to eventually address https://github.com/newrelic/terraform-provider-newrelic/issues/839.

Personally, I'm a bit confused how this client API prefers to implement the NerdGraph APIs.

### DIFF
--- a/pkg/apiaccesskeys/apiaccesskey.go
+++ b/pkg/apiaccesskeys/apiaccesskey.go
@@ -1,0 +1,234 @@
+package apiaccesskeys
+
+import (
+	"errors"
+	"fmt"
+	"github.com/newrelic/newrelic-client-go/internal/http"
+)
+
+// APIAccessKey represents a New Relic API access ingest or user key.
+type APIAccessKey struct {
+	ID         string `json:"id,omitempty"`
+	Key        string `json:"key,omitempty"`
+	Name       string `json:"name,omitempty"`
+	Notes      string `json:"notes,omitempty"`
+	Type       string `json:"type,omitempty"`
+	AccountID  int    `json:"accountId,omitempty"`
+	IngestType string `json:"ingestType,omitempty"`
+	UserID     string `json:"userId,omitempty"`
+}
+
+// apiAccessKeyCreateResponse represents the JSON response returned from creating key(s).
+type apiAccessKeyCreateResponse struct {
+	APIAccessCreateKeys struct {
+		CreatedKeys []APIAccessKey                    `json:"createdKeys,omitempty"`
+		Errors      []apiAccessKeyMutationErrResponse `json:"errors,omitempty"`
+	} `json:"apiAccessCreateKeys"`
+}
+
+// apiAccessKeyUpdateResponse represents the JSON response returned from updating key(s).
+type apiAccessKeyUpdateResponse struct {
+	APIAccessUpdateKeys struct {
+		UpdatedKeys []APIAccessKey                    `json:"updatedKeys,omitempty"`
+		Errors      []apiAccessKeyMutationErrResponse `json:"errors,omitempty"`
+	} `json:"apiAccessUpdateKeys"`
+}
+
+// apiAccessKeyGetResponse represents the JSON response returned from getting an access key.
+type apiAccessKeyGetResponse struct {
+	Actor struct {
+		APIAccess struct {
+			Key APIAccessKey `json:"key,omitempty"`
+		} `json:"apiAccess"`
+	} `json:"actor"`
+	http.GraphQLErrorResponse
+}
+
+// apiAccessKeyMutationErrResponse represents the generic error response returned from modifying API access keys.
+type apiAccessKeyMutationErrResponse struct {
+	AccountID       int    `json:"accountId,omitempty"`
+	ID              string `json:"id,omitempty"`
+	IngestErrorType string `json:"ingestErrorType,omitempty"`
+	UserErrorType   string `json:"userErrorType,omitempty"`
+	IngestType      string `json:"ingestType,omitempty"`
+	Message         string `json:"message,omitempty"`
+	Type            string `json:"type,omitempty"`
+}
+
+// apiAccessKeyDeleteResponse represents the JSON response returned from creating key(s).
+type apiAccessKeyDeleteResponse struct {
+	APIAccessDeleteKeys struct {
+		DeletedKey []struct {
+			ID string `json:"id"`
+		} `json:"deletedKeys,omitempty"`
+		Errors []apiAccessKeyMutationErrResponse `json:"errors,omitempty"`
+	} `json:"apiAccessDeleteKeys"`
+}
+
+const (
+	graphqlAPIAccessKeyBaseFields = `
+		id
+		key
+		name
+		notes
+		type
+		... on ApiAccessIngestKey {
+			id
+			name
+			accountId
+			ingestType
+			key
+			notes
+			type
+		}
+		... on ApiAccessUserKey {
+			id
+			name
+			accountId
+			key
+			notes
+			type
+			userId
+		}
+		... on ApiAccessKey {
+			id
+			name
+			key
+			notes
+			type
+		}`
+
+	graphqlAPIAccessCreateKeyFields = `createdKeys {` + graphqlAPIAccessKeyBaseFields + `}`
+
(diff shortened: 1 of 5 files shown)


pkg/apiaccesskeys/apiaccesskey.go
```diff
@@ -0,0 +1,103 @@
...
+	resp := apiAccessIngestKeyCreateResponse{}
+
+	if err := a.client.NerdGraphQuery(apiAccessIngestKeyCreateKeys, vars, &resp); err != nil {
+		return nil, err
+	}
+
+	return resp.APIAccessKeys.CreatedKeys, nil
+}
+
+// UpdateAPIAccessKeyMutation
```
sanderblue, line 88: `UpdateAPIAccessKeyMutation` still a WIP? 🙂

pkg/apiaccesskeys/apiaccesskey.go
```diff
@@ -0,0 +1,103 @@
...
+				}
+				errors {
+					message
+					type
+				}
+		}}`
+)
+
+// CreateAPIAccessIngestKeyMutation creates a new ingest type API access key.
+func (a *APIAccessKeys) CreateAPIAccessIngestKeyMutation(keys APIAccessCreateIngestKeysInput) ([]APIAccessIngestKey, error) {
```
sanderblue, line 74: This is great. Similar comment to earlier, we should consider that this mutation facilitates creating **ingest keys** _and_ **user keys**. So maybe the API looks something like this: 

```suggestion
func (a *APIAccessKeys) CreateAPIAccessKeysMutation(keys APIAccessCreateKeysInput) ([]APIAccessKey, error) {
```

This would facilitate creating both types of keys and make the client's API a 1:1 mapping to the GraphQL mutation signature 😎

pkg/apiaccesskeys/apiaccesskey.go
```diff
@@ -0,0 +1,103 @@
...
+// APIAccessUserKey represents a New Relic API access user key.
+type APIAccessUserKey struct {
+	APIAccessKeyBase
+	AccountID int    `json:"accountId,omitempty"`
+	UserID    string `json:"userId,omitempty"`
+}
+
+type apiAccessIngestKeyCreateResponse struct {
+	APIAccessKeys struct {
+		CreatedKeys []APIAccessIngestKey `json:"createdKeys"`
```
sanderblue, line 28: One thing we should consider is that the create mutation facilitates creating **ingest keys** _and_ **user keys**. We might want to think about adjusting this type definition to allow that.

e.g. `CreatedKeys []APIAccessKey` -- this could return both user keys and ingest keys.

sanderblue:
One other thing that's preventing the `lint` stage from passing in our ci pipeline is the commit message format. We use "conventional commit" formatting for our commit messages. Here is a bit more information on [how we format our commit messages](https://github.com/newrelic/newrelic-client-go#commit-messages). Do you mind amending your commit messages? They will look something like the following:

```
feat(apiaccesskeys): add new api access keys package 
```

pkg/apiaccesskeys/apiaccesskey.go
davidji99: Yep =)

davidji99:
> One other thing that's preventing the `lint` stage from passing in our ci pipeline is the commit message format. We use "conventional commit" formatting for our commit messages. Here is a bit more information on [how we format our commit messages](https://github.com/newrelic/newrelic-client-go#commit-messages). Do you mind amending your commit messages? They will look something like the following:
> 
> ```
> feat(apiaccesskeys): add new api access keys package 
> ```

I'll squash commits once the PR is ready to go per the commit message format you mentioned above.

pkg/apiaccesskeys/apiaccesskey.go
```diff
@@ -0,0 +1,103 @@
...
+		name
+		notes
+		type
+		... on ApiAccessIngestKey {
+			ingestType
+			accountId
+		}
+	`
+
+	apiAccessIngestKeyCreateKeys = `mutation($keys: Int!){
```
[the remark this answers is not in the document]
davidji99: Ah, thanks for this. I didn't know how to construct the mutation argument type.

pkg/apiaccesskeys/apiaccesskey.go
```diff
@@ -0,0 +1,103 @@
...
+			apiAccessCreateKeys(keys: $keys) {createdKeys {` + graphqlAPIAccessIngestKeyFields + `}
+		}}`
+
+	apiAccessIngestKeyDeleteKey = `mutation($keys: Int!){
```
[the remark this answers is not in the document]
davidji99: Roger.

pkg/apiaccesskeys/apiaccesskey.go
davidji99: Ah I was hoping to make it easier for users to use specific methods named after the key type. Also, my motivation to work on this PR is for ingest keys only.

I can explore a consolidated approach but if it proves a bit too much for me, I will likely back out and implement these endpoints on my fork.

pkg/apiaccesskeys/apiaccesskey.go
```diff
@@ -0,0 +1,103 @@
...
+			apiAccessDeleteKeys(keys: $keys) {
+				deletedKeys {
+					id
+				}
+				errors {
+					message
+					type
```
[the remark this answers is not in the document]
davidji99: If we return this, I think we would need some sort of helper method to parse and detect there are errors returned in the response. I noticed that `*.client.NerdGraphQuery` does not return an appropriate HTTP code if there is an error?

davidji99:
Hi @sanderblue,

Thank you very much for your initial review. I've implemented the CRUD endpoints for API Access keys. Could you review again?

Also does the integration test framework in this library require valid authentication tokens? I ask as I got this error when running a very basic test:
```
    require.go:360: 
        	Error Trace:	require.go:360
        	            				apiaccesskeys_integration_test.go:29
        	Error:      	Expected nil, but got: &errors.errorString{s:"{1234567  FORBIDDEN  BROWSER You do not have permission to create this key. INGEST}\n"}
        	Test:       	TestIntegrationAPIAccessKeys
```

pkg/apiaccesskeys/apiaccesskey.go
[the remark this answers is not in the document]
sanderblue: The client's intent is to be more-or-less a pass-through and bubble up any API errors that may occur, and currently the API has a few inconsistencies with how errors are presented and formatted in the raw response. The API team is working on standardizing on a solid pattern, but in the meantime, you're correct, we might need a helper function for error capture/formatting.

sanderblue:
Hi @davidji99, thanks for making those updates. Everything is looking great so far. I'm taking a closer look, but we're definitely very close to this being ready to go. 

Regarding unit and integration tests, it looks like the actual tests are passing, but our custom JUnit reporter step is failing, which is just fine. The custom reporter is there to provide metadata about our test runs, but it looks like we need to make an adjustment so it doesn't make it look like the pipeline failed. Thanks for mentioning this.

davidji99:
> Awesome contribution @davidji99! Thanks for taking the time to add this feature :shipit:

No problem! Thanks @sanderblue for the review and assistance.

Could you make a new release so I can start on the TF provider changes?

sanderblue:
Yessir, we'll try to get a release out before the end of tomorrow. I have a fast-follow branch for this feature that we'll want to merge before releasing (we have a code generator tool that we use to generate types for our packages based on the GraphQL types, which will lead to some very minor adjustments). Our contribution guide doesn't contain this information yet because the tool we use is still in the early beta stages of development. But either way, the bulk of this code will live on, it'll mostly just be some type updates 🙂

davidji99:
Ah gotcha and sounds good. Yeah, I noticed that the types were auto-generated but I didn't know how to run that script.

davidji99:
Hi @sanderblue,

Since you're going to do a follow-up PR, could change this field type from `String` to `Int` https://github.com/newrelic/newrelic-client-go/blob/master/pkg/apiaccesskeys/apiaccesskey.go#L18?

sanderblue:
Will take a look for sure. Thanks for mentioning it :)

PR #433, opened 2020-08-19
chore(apiaccess): domain adjustments for apiaccess resource

sanderblue:
This PR builds on @davidji99's pull request (#429  👏 ) which added the `apiaccess` resource. This PR primarily involves domain-specific updates pertaining to our code generation CLI ([`tutone`](https://github.com/newrelic/tutone)) process of our Go types that get generated based on the GraphQL schema types introspection.

**TL;DR**

- Updates `apiaccesskeys` package name to `apiaccess` for a 1:1 mapping to the top-level NerdGraph `apiAccess` entry points (this also includes updating type names to keep things as consistent as possible)
- Updates the associated `types.go` file with the generated types based on the NerdGraph schema using `tutone`
- Updates the `APIAccess` CRUD methods due to types being updated via `tutone` code generation

Again, big thanks to @davidji99!


davidji99:
Hi @sanderblue,

I have a few questions:
1. For `func (a *APIAccess) GetAPIAccessKey(keyID string, keyType ApiAccessKeyType) (*ApiAccessKey, error)`, the NerdGraph Query (`apiAccessKeyGetKey`) specifies `AccountId` as a returned field but the `ApiAccessKey` struct does not define this field. Is there an alternative method to use that returns the base Key fields along with the key type specific fields? My original PR defined [all possible fields](https://github.com/newrelic/newrelic-client-go/pull/433/files#diff-d0899ddc84561135f98eca75d6970eb8L10) which worked out well for the provider pull request. The same goes for `ingestType` and `UserID` field. Is there another way to get additional fields?
1. I've noticed `.UserId` isn't `.UserID`. Is the tutone CLI not formatting 'ID' correctly?
1. It would be good for `type ApiAccessKeyType string` and `type ApiAccessUserKeyErrorType string` to implement a `.String()` method so users don't need to do `string(...)` for type conversion. For example, I need to use strings to put in `[]string{}` variable.

sanderblue:
Hi @davidji99,

Thanks for the review! You've brought up some valid points.

> 1. For `func (a *APIAccess) GetAPIAccessKey(keyID string, keyType ApiAccessKeyType) (*ApiAccessKey, error)`, the NerdGraph Query (`apiAccessKeyGetKey`) specifies `AccountId` as a returned field but the `ApiAccessKey` struct does not define this field. Is there an alternative method to use that returns the base Key fields along with the key type specific fields? My original PR defined [all possible fields](https://github.com/newrelic/newrelic-client-go/pull/433/files#diff-d0899ddc84561135f98eca75d6970eb8L10) which worked out well. The same goes for `ingestType` and `UserID` field. Is there another way to get these missing fields?
> 2. I've noticed `.UserId` isn't `.UserID`. Is the tutone CLI not formatting 'ID' correctly?
> 3. It would be good for `type ApiAccessKeyType string` and `type ApiAccessUserKeyErrorType string` to implement a `.String()` method so users don't need to do `string(...)` for type conversion. For example, I need to use strings to put in `[]string{}` variable.

1. Good catch. The types for this were honestly a little tricky with regard to code generation, so I appreciate the detailed look. It looks like I need to stitch together a return type like you did to meet those requirements. Thanks for the patience.

2. I noticed this as well. Same with `Api` vs `API`. I thought we ran `gofmt` and used some other tooling on the output file, but I'll revisit and will get this updated to be formatted in the normal Go standards that we've been following.

3. Fair enough. We've often wrapped those values with `string()` in our own code, but a helper method on the struct seems useful here.

sanderblue:
@davidji99 can you take a look at the latest updates? I wasn't able to get to adding the String() methods to those types due to a lot of other things going on at the moment, but I think the other items should be addressed.

pkg/apiaccess/types.go
```diff
@@ -0,0 +1,245 @@
...
+	// The notes you want to assign to the key.
+	Notes string `json:"notes"`
+}
+
+// ID - The `ID` scalar type represents a unique identifier, often used to
+// refetch an object or as key for a cache. The ID type appears in a JSON
+// response as a String; however, it is not intended to be human-readable.
+// When expected as an input type, any string (such as `"4"`) or integer
+// (such as `4`) input value will be accepted as an ID.
+type ID string
```
zlesnr, line 235: This probably isn't needed.  I think the alerts package for the tutone config has an example for how to ignore it.

pkg/apiaccess/types.go
```diff
@@ -0,0 +1,245 @@
...
+
+// APIAccessKeyError - A key error. Each error maps to a single key input.
+type APIAccessKeyErrorInterface interface {
```
zlesnr, line 238: Nice, tutone PR inbound I take it for the addition of `Interface` here yeah?  I like it.

pkg/apiaccess/types.go
sanderblue: Cool, I'll take a look. I was following the previously added package I think.

pkg/apiaccess/types.go
sanderblue: Yeah, with several other little updates 😎

pkg/apiaccess/types.go
sanderblue: I removed this and updated the tutone config for this package so it doesn't get created :)

davidji99:
Hi @sanderblue,

Latest PR changes work great for https://github.com/newrelic/terraform-provider-newrelic/pull/850. Thanks for the changes!

sanderblue:
@davidji99 thanks for confirming, I'll get a release going shortly :shipit:

PR #438, opened 2020-08-25
feat(logs): Implement Log API

msummers-nr:
Cleaned-up based on the previous comments and fixed the commit message.


pkg/logs/example_basic_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"log"
+	"os"
+
+	"github.com/newrelic/newrelic-client-go/pkg/config"
+)
+
+//func TestExample_basic(t *testing.T) {
+func Example_basic() {
+	// Initialize the client configuration.  An Insights insert key is required
```
ctrombley, line 14: nit: s/Insights insert key/New Relic license key/

pkg/logs/logs.go
```diff
@@ -0,0 +1,53 @@
...
+		return errors.New("logs: CreateLogEntry: logEntry is nil, nothing to do")
+	}
+
+	rsp, err := l.client.Post(l.config.Region().LogsURL(), nil, logEntry, nil)
+
+	if err != nil {
+		return err
+	}
+
+	if !(rsp.StatusCode < 299) {
```
ctrombley, line 48: +1 on @sanderblue 's comment from earlier.  This case should be covered from within the `Post` method, and so this check seems redundant.  By the time we get here, this conditional will always evaluate to true unless `rsp.StatusCode == 299`.

msummers-nr: For a reason I could not identify that code did not work, this was the only way to get it to succeed.

pkg/logs/example_basic_test.go
msummers-nr: Good catch

pkg/logs/logs.go
ctrombley: I'm interested to dig a little deeper on this in case we have some bugs hidden in the client.  Do you have a repro case?  The integration test provided in this PR seems to succeed with or without this check.

That being said, we don't need to hold this PR up while I look into this.

msummers-nr: No, I don't. We'd have to rewrite that bit of code an try it out.

PR #447, opened 2020-09-02
feat(nrql_conditions): add new fields 'expiration' and 'signal'

sanderblue:
Facilitates: [newrelic/terraform-provider-newrelic #810](https://github.com/newrelic/terraform-provider-newrelic/issues/810)

### DIFF
--- a/pkg/alerts/nrql_conditions.go
+++ b/pkg/alerts/nrql_conditions.go
@@ -10,7 +10,7 @@ import (
 // Settings for how violations are opened or closed when a signal expires.
 type AlertsNrqlConditionExpiration struct {
 	CloseViolationsOnExpiration bool `json:"closeViolationsOnExpiration"`
-	ExpirationDuration          int  `json:"expirationDuration"`
+	ExpirationDuration          *int `json:"expirationDuration,omitempty"`
 	OpenViolationOnExpiration   bool `json:"openViolationOnExpiration"`
 }
 
@@ -18,7 +18,7 @@ type AlertsNrqlConditionExpiration struct {
 type AlertsNrqlConditionSignal struct {
 	EvaluationOffset int              `json:"evaluationOffset"`
 	FillOption       AlertsFillOption `json:"fillOption"`
-	FillValue        float64          `json:"fillValue"`
+	FillValue        *float64         `json:"fillValue,omitempty"`
 }
 
 // AlertsFillOption - The available fill options.
--- a/pkg/alerts/nrql_conditions.go
+++ b/pkg/alerts/nrql_conditions.go
@@ -584,36 +584,36 @@ type getNrqlConditionQueryResponse struct {
 
 const (
 	graphqlNrqlConditionStructFields = `
-		id
-		name
-		nrql {
-			evaluationOffset
-			query
-		}
-		enabled
-		description
-		policyId
-		runbookUrl
-		terms {
-			operator
-			priority
-			threshold
-			thresholdDuration
-			thresholdOccurrences
-		}
-		type
-		violationTimeLimit
-		expiration {
+    id
+    name
+    nrql {
+      evaluationOffset
+      query
+    }
+    enabled
+    description
+    policyId
+    runbookUrl
+    terms {
+      operator
+      priority
+      threshold
+      thresholdDuration
+      thresholdOccurrences
+    }
+    type
+    violationTimeLimit
+    expiration {
       closeViolationsOnExpiration
       expirationDuration
       openViolationOnExpiration
-		}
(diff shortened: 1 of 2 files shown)


pkg/alerts/nrql_conditions.go
```diff
@@ -6,6 +6,40 @@ import (
...
+	CloseViolationsOnExpiration bool `json:"closeViolationsOnExpiration"`
+	ExpirationDuration          int  `json:"expirationDuration"`
+	OpenViolationOnExpiration   bool `json:"openViolationOnExpiration"`
+}
+
+// AlertsNrqlConditionSignal - Configuration that defines the signal that the NRQL condition will use to evaluate.
+type AlertsNrqlConditionSignal struct {
+	EvaluationOffset int              `json:"evaluationOffset"`
+	FillOption       AlertsFillOption `json:"fillOption"`
+	FillValue        float64          `json:"fillValue"`
```
ahornerr, line 21: Fill value is nullable in the API, suggest making it a pointer

pkg/alerts/nrql_conditions.go
```diff
@@ -6,6 +6,40 @@ import (
 	"github.com/newrelic/newrelic-client-go/pkg/errors"
 )
 
+// AlertsNrqlConditionExpiration - **Preview access:** These fields may be viewed and set, but will not be active until the release date.
+// Settings for how violations are opened or closed when a signal expires.
+type AlertsNrqlConditionExpiration struct {
+	CloseViolationsOnExpiration bool `json:"closeViolationsOnExpiration"`
+	ExpirationDuration          int  `json:"expirationDuration"`
```
ahornerr, line 13: Expiration duration is nullable in the API, suggest making it a pointer

pkg/alerts/nrql_conditions.go
sanderblue: roger that

pkg/alerts/nrql_conditions.go
sanderblue: 10-4

ahornerr (dismissed):

ahornerr (dismissed):

pkg/alerts/nrql_conditions.go
```diff
@@ -6,6 +6,42 @@ import (
 	"github.com/newrelic/newrelic-client-go/pkg/errors"
 )
 
+// AlertsNrqlConditionExpiration - **Preview access:** These fields may be viewed and set, but will not be active until the release date.
+// Settings for how violations are opened or closed when a signal expires.
+// nolint:golint
+type AlertsNrqlConditionExpiration struct {
+	ExpirationDuration          *int `json:"expirationDuration"`
```
zlesnr, line 13: `omitempty`?

zlesnr: Same question for the other pointers in `AlertsNrqlConditionSignal`.

zlesnr (commented):
This looks good to me.  Just a question about when to include the `omitempty` on the pointer fields.

pkg/alerts/nrql_conditions.go
```diff
@@ -565,6 +603,16 @@ const (
 		}
 		type
 		violationTimeLimit
+		expiration {
+      closeViolationsOnExpiration
```
ctrombley, line 607: nit: fix indentation?

ctrombley (approved):
🌮

pkg/alerts/nrql_conditions.go
ahornerr: The nuance of the GraphQL service, merging of previous/new fields, optional values, default values in the API, and backwards compatible fields requires a somewhat specific combination of pointers and inclusion/exclusion of `omitempty`.

For the most part, these have to be pointers in order to allow the Terraform to be backwards compatible (not requiring users to begin setting these options until they're ready). GraphQL differentiates between nil with `omitempty` (Javascript `undefined`) versus nil without `omitempty` (Javascript `null`). In order to support options like `fillValue`, it is required some fields must have the ability to be null.

pkg/alerts/nrql_conditions.go
sanderblue: Not sure I see what you're seeing 🤔

sanderblue: Ah, it's my chrome extension thats hiding it. Must be a copy and paste thing.

zlesnr (dismissed):

zlesnr:
🐐 ☁️

zlesnr (approved):
More 🐐

PR #452, opened 2020-09-02
feat(application_instances): add an application instance resource

### LINKED ISSUE #444: Support retrieval of application instances
### AC:
- Support the following REST endpoints:
  - https://rpm.newrelic.com/api/explore/application_instances/list
  - https://rpm.newrelic.com/api/explore/application_instances/show

ctrombley:
Resolves #444 .

This PR adds methods for retrieving application instances.


pkg/apm/application_instances_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	testApplicationInstanceLinks = ApplicationInstanceLinks{
```
sanderblue, line 13: Not a blocker, but these variables could be moved into the `TestListApplicationInstancesWithParams` function scope to avoid variable name collisions across the test suite.

sanderblue (dismissed):
🌮

pkg/apm/application_instances_test.go
ctrombley: 👍

pkg/apm/application_instances.go
```diff
@@ -0,0 +1,74 @@
...
+
+// ApplicationInstanceLinks represents all the links for a New Relic application instance.
+type ApplicationInstanceLinks struct {
+	Application     int `json:"application,omitempty"`
+	ApplicationHost int `json:"application_host,omitempty"`
+}
+
+// ApplicationInstance represents information about a New Relic application instance.
+type ApplicationInstance struct {
+	ID              int                       `json:"id,omitempty"`
```
sanderblue, line 13: Not a huge deal in all likelihood, but would we want to remove `omitempty` for an ID in case an ID is zero?

sanderblue: I guess in general I'm wondering about usage of `omitempty` in general for integers than could be valid values of `0`.

ctrombley: This is a read-only resource, so it's probably a non-issue?

sanderblue: Si, de acuerdo

sanderblue (approved):
🌮

PR #461, opened 2020-09-09
Allow logs to use insert key or license key

bpeckNR:
Allowing logs to use an insert key or a license key. The Logs API expects an insert key to be added with the header Api-Key, which is a bit different than the X-Insert-Key header typically used for insert keys. This changes auth.go to include an additional mechanism and changes logs.go to allow users to choose license key or insert key based authorization based on what options are set in the config.


bpeckNR:
A new PR will be submitted that follows contributing guidelines.

PR #475, opened 2020-09-15
fix(http): handle 'not found' downstream response

sanderblue:
Facilitates resolving: [terraform-provider-newrelic #860](https://github.com/newrelic/terraform-provider-newrelic/issues/860)

The issue above might have been caused by an upstream API change 🤔 💭


zlesnr (approved):
Nice work.  🐐

PR #477, opened 2020-09-15
fix: adjustments for 'Not Found' scenarios

sanderblue:
NerdGraph's API response seems to have changed for the alert policy query.

**Old API Behavior:** Querying for a non-existent alert policy would return a response containing error field values of `BAD_USER_INPUT` and `Not Found`.
**New API Behavior:** The response now returns a `null` policy if a policy is not found. This breaks things.


mattbaker:
This has been rolled back. We'll work with ya to migrate.

PR #478, opened 2020-09-16
refactor(alerts): remove widespread change, limit scope to only nrql condition error resp handling

sanderblue:
Refactors #475 

The previous change was a widespread change and currently it's safer to scope this change to the resource in question rather than for all of them.

### DIFF
--- a/pkg/alerts/nrql_conditions.go
+++ b/pkg/alerts/nrql_conditions.go
@@ -2,7 +2,9 @@ package alerts
 
 import (
 	"fmt"
+	"strings"
 
+	"github.com/newrelic/newrelic-client-go/internal/http"
 	"github.com/newrelic/newrelic-client-go/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/alerts/nrql_conditions.go
```diff
@@ -353,7 +355,15 @@ func (a *Alerts) GetNrqlConditionQuery(
 		"id":        conditionID,
 	}
 
-	if err := a.client.NerdGraphQuery(getNrqlConditionQuery, vars, &resp); err != nil {
+	req, err := a.client.NewNerdGraphRequest(getNrqlConditionQuery, vars, &resp)
```
zlesnr, line 358: Has the signature changed here?

sanderblue: Nope, it's just a different way of performing a request. `NerdGraphQuery` uses `NewNerdGraphRequest`, so in order to set the request's error response, we need to create the request in a more granular way. We did this in the alert policy resource as well 😎 

I also think it would be a bit cleaner to have an additional argument passed to  `NerdGraphQuery ` ( e.g. `errorResponse`) and then set the errorResponse within that method, but would affect all the things.

zlesnr (approved):

PR #489, opened 2020-09-30
fix(events): dereference the data pointer

zlesnr:
Without the change, the received data type is a pointer and never posted to
insights.  Here we dereference the pointer so the data flows properly.


zlesnr:
This should resolve https://github.com/newrelic/newrelic-cli/issues/385

sanderblue (approved):
🐑 it

PR #493, opened 2020-10-02
feat: remove admin API key as an authentication mechanism

ctrombley:
This PR removes Admin API keys as an authentication mechanism.


sanderblue (approved):
🐑 it

PR #514, opened 2020-10-16
feat(http): retry on nerdgraph server timeout

### LINKED ISSUE #488: Handle timeout errors with a retry
```
2020-09-30T21:14:04.526Z [DEBUG] plugin.terraform-provider-newrelic_v2.7.5: time="2020-09-30T21:14:04Z" level=trace msg="request completed" body="{\"data\":{\"alertsNrqlConditionStaticCreate\":null},\"errors\":[{\"extensions\":{\"errorClass\":\"TIMEOUT\"},\"locations\":[{\"column\":4,\"line\":3}],\"message\":\"Resolution of this value experienced a timeout.\",\"path\":[\"alertsNrqlConditionStaticCreate\"]}]}" headers="{\"Cache-Control\":[\"max-age=0, private, must-revalidate\"],\"Content-Security-Policy\":[\"frame-ancestors *.newrelic.com\"],\"Content-Type\":[\"application/json; charset=utf-8\"],\"Date\":[\"Wed, 30 Sep 2020 21:14:04 GMT\"],\"Proxied-By\":[\"Service Gateway\"],\"Served-By\":[\"nerd-graph\"],\"Server\":[\"Cowboy\"],\"Strict-Transport-Security\":[\"max-age=31536000; includeSubDomains\"]}" method=POST status_code=200 url="https://staging-api.newrelic.com/graphql"
```

We should be able to retry in the event we receive an error message back from nerdgraph that has the TIMEOUT in the body.

ctrombley:
Resolves #488 .

This PR allows the client to retry requests on NerdGraph server timeouts.  It leaves go-retryablehttp in place, but makes use of its retry configuration and helper methods to do a second pass of retry logic after the response body has been read.  The end result is a move away from cohesion in the retry logic (and potential for more than `RetryMax` retries in rare cases), but we will accept this in light of some ongoing impactful customer issues.  We will want to make a refactor pass at a later date to better factor the retry logic and potentially replace go-retryablehttp entirely.

cc @abuchanan-nr

### DIFF
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -249,100 +249,27 @@ func logCleanHeaderMarshalJSON(header http.Header) ([]byte, error) {
 }
 
 // Do initiates an HTTP request as configured by the passed Request struct.
-// nolint:gocyclo
 func (c *Client) Do(req *Request) (*http.Response, error) {
-	r, err := req.makeRequest()
-	if err != nil {
-		return nil, err
-	}
+	var resp *http.Response
+	var errorValue ErrorResponse
+	var body []byte
 
 	logger := c.config.GetLogger()
-
-	logger.Debug("performing request", "method", req.method, "url", r.URL)
-
-	logHeaders, err := logCleanHeaderMarshalJSON(r.Header)
-	if err != nil {
-		return nil, err
-	}
-
-	if req.reqBody != nil {
-		switch reflect.TypeOf(req.reqBody).String() {
-		case "*http.graphQLRequest":
-			x := req.reqBody.(*graphQLRequest)
-
-			logVariables, marshalErr := json.Marshal(x.Variables)
-			if marshalErr != nil {
-				return nil, marshalErr
-			}
-
-			logger.Trace("request details",
-				"headers", logNice(string(logHeaders)),
-				"query", logNice(x.Query),
-				"variables", string(logVariables),
-			)
-		case "string":
-			logger.Trace("request details", "headers", string(logHeaders), "body", logNice(req.reqBody.(string)))
-		}
-	} else {
-		logger.Trace("request details", "headers", string(logHeaders))
-	}
-
-	var shouldRetry bool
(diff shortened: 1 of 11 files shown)


internal/http/client.go
```diff
@@ -286,31 +287,63 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
...
+		}
+
+		logHeaders, err = json.Marshal(resp.Header)
+		if err != nil {
+			return nil, err
+		}
 
-	errorValue := req.errorValue.New()
-	_ = json.Unmarshal(body, &errorValue)
+		logger.Trace("request completed", "method", req.method, "url", r.URL, "status_code", resp.StatusCode, "headers", string(logHeaders), "body", string(body))
```
buchanae, line 325: Maybe include an `attempt` attribute? So we can see that it is retrying.

edit: oh, I see the "retrying" debug line above. attempt still might be nice.

internal/http/client.go
```diff
@@ -286,31 +287,63 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
...
+		resp, retryErr = c.client.Do(r)
+		if retryErr != nil {
+			return nil, retryErr
+		}
 
-	logHeaders, err = json.Marshal(resp.Header)
-	if err != nil {
-		return nil, err
-	}
+		defer resp.Body.Close()
```
buchanae, line 309: I think these won't be closed until all retries are finished. That's probably ok, but suboptimal. I usually put the body of my retry loop in a separate function so that I can use defer naturally.

buchanae (commented):
Nice!

internal/http/client.go
ctrombley: Agree. Updated 👍

internal/http/client.go
ctrombley: Updated 👍

sanderblue (approved):
🌮

PR #529, opened 2020-10-28
fix(apm): apm.MetricTimesliceValues is not a hashmap

kminehart:
According to [the API explorer](https://rpm.newrelic.com/api/explore/applications/metric_data), APM metric data is a hashmap and not a struct with known fields. Some fields that we would like to use are missing unfortunately.

Just so that this change doesn't break existing clients, I've added a values map on on the `apm.MetricTimesliceValues` struct, and a custom unmarshal function that will populate this map as well as the struct fields, and a test that should represent that change.


kminehart:
```

=== Errors
pkg/apm/applications_test.go:315:5: too few values in MetricTimesliceValues literal
pkg/apm/applications_test.go:327:5: too few values in MetricTimesliceValues literal
```

What exactly should I do about this?

The literals in-question:

```go
expect := &MetricTimesliceValues{
→ AsPercentage:           10.3,
→ AverageTime:            133.1,
→ CallsPerMinute:         3029.409,
→ MaxValue:               500.0,
→ TotalCallTimePerMinute: 123.122,
→ Utilization:            0.0,
→ Values: map[string]float64{
→ → "as_percentage":              10.3,
→ → "average_time":               133.1,
→ → "calls_per_minute":           3029.409,
→ → "max_value":                  500.0,
→ → "total_call_time_per_minute": 123.122,
→ → "standard_deviation":         100.155,
→ → "other_key":                  22.11,
→ },
}

values := MetricTimesliceValues{}
```

ctrombley:
Thanks @kminehart !  This is a pretty big oversight and we are really happy to have the contribution.  I was going to ask if you could rebase your commit messages to match the guidance in the contribution guide, but then I realized that section was missing from the guide in this repo.  I've updated it now and you can find it [here](https://github.com/newrelic/newrelic-client-go/blob/master/CONTRIBUTING.md#commit-messages).  We use the commit messages to generate changelogs and therefore we use a linter at build time to check them all - I think that's why the build is failing here.

kminehart:
ah. I assumed that was related to the PR title and I was pulling my hair out trying to figure it out. I will rebase the messages to include `fix(apm): {message}`

kminehart:
Everything should be updated, I think I saw the lint action pass a second ago.

ctrombley (approved):
⚡

ctrombley:
Fantastic @kminehart, I'll cut a release shortly.

PR #531, opened 2020-10-29
chore(edge): update generated code and include methods

zlesnr:


zlesnr:
Requires https://github.com/newrelic/tutone/pull/99

PR #540, opened 2020-11-09
fix(http): include INTERNAL_SERVER_ERROR as a retry reason

zlesnr:
Without this change, a downstream error returned by GraphQL of
INTERNAL_SERVER_ERROR is not included in the reasons to retry an call.  Here we
rename the IsTimeout method to IsRetryableError to be a bit more generic, and
include another condition in which we might want to retry the call.


sanderblue (approved):
🐑 🐑 🐑 🐑 🐑 🐑 🐑 🐑  it

🦙

zlesnr:
Test failures look to be upstream API failures.  Merging.

PR #553, opened 2020-11-24
Allow for paging when listing monitors

nickschuch:
Currently Synthetics `ListMonitors()` is hardcoded to 100 items.

This pull request:
* Move params to be configurable
* Adds fields for offset handling
* Move response for offset field


ctrombley:
Hi @nickschuch!  Passing a nil `queryParams` will cause a panic.  Can you allow for a default limit to be used when passing nil, and add a unit test to support that use case?

jthurman42:
This has been resolved in #705

PR #560, opened 2020-12-01
Updating from current NerdGraph Schema

jthurman42:
Updating the schema has some breaking changes, mostly around capitalization of fields in the schema.  At least:

- `Nrql` to `NRQL`
- `Nrdb` to `NRDB`

Note, this requires newrelic/tutone#116 as the template was updated for those modifications.


templates/typegen/types.go.tmpl
```diff
@@ -16,13 +16,17 @@ type {{ .Name }} string
...
+  {{-   if ne .Description "" }}
+  {{      .Description }}
+  {{-   end }}
+  {{   .Name }} {{ $typeName }}
   {{- end}}
 }{
   {{- range .Values }}
-	{{ .Description }}
-	{{ .Name }}: "{{ .Name }}",
+  {{-   if ne .Description "" }}
```
zlesnr, line 26: I think `{{ if .Description }}` would also work in templates.

zlesnr (approved):

PR #575, opened 2020-12-16
Code Generate Entities

jthurman42:
Background work for Dashboards, as Dashboards are Entities, so fetching them is via this package

### Queries:
* actor.entities
* actor.entity
* actor.entitySearch

*Note:* Tag queries are not auto-generated as they are under `actor.entity` which Tutone does not currently support nested endpoints.

### Mutations:

* TaggingAddTagsToEntity (Deprecates `AddTags`)
* TaggingDeleteTagFromEntity (Deprecates `DeleteTag`)
* TaggingDeleteTagValuesFromEntity (Deprecates `DeleteTagValues`)
* TaggingReplaceTagsOnEntity (Deprecates `ReplaceTags`)

### Other:
* Fix template so you can Unmarshal types that were overridden from another package
* DRY up Nrdb from Entities


.tutone.yml
```diff
@@ -183,49 +184,147 @@ packages:
...
+      - name: MobileApplicationEntity
+      - name: SyntheticMonitorEntity
+      - name: ThirdPartyServiceEntity
+      - name: UnavailableEntity
+      - name: UnavailableEntityOutline
+      - name: WorkloadEntity
+
+      #
+      # Types that our other packages define
+      #   TODO: Fix tutone to figure this out for us
```
zlesnr, line 266: Maybe ticket this in place of a comment here.

.tutone.yml
```diff
@@ -183,49 +184,147 @@ packages:
...
+        max_query_field_depth: 1
+
     types:
       - name: AttributeMap
         create_as: "map[string]interface{}"
-#      - name: Entity
-#      - name: EntityOutline
+
+      # Types this package owns
+      - name: AlertableEntity
```
zlesnr, line 224: Since you have the `queries` above set for this package, the expander should drag in all these types without having to specify them manually here.  Are they here just for posterity?

templates/typegen/types.go.tmpl
```diff
@@ -73,7 +77,12 @@ func (x *{{ $typeName }}) UnmarshalJSON(b []byte) error {
         }
 
         for _, m := range rawMessage{{ .Name }} {
+          {{- if contains "." $field.TypeName }}
+          {{-   $m := split "." $field.TypeName }}
+          xxx, err := {{ $m._0 }}.Unmarshal{{ $m._1 }}Interface(*m)
```
zlesnr, line 82: What is `$m._0` doing here?

zlesnr (approved):

.tutone.yml
jthurman42: Basically?  One item I've been thinking a lot about is a way for Tutone to understand cross-package deps.  The initial thought was if you statically define the type, then that package "Owns" it, and therefore if required by another package, tutone could import/auto-reference the other package and the type.  None of that has been done in Tutone, so having them here is not explicitly required, but somewhat serves a manual purpose of claiming them...

templates/typegen/types.go.tmpl
jthurman42: Template language for `m[0]`.  This resolves referencing interface unmarshallers defined in another package.

Previously generated: `xxx, err := Unmarhsalpackage.TypeInterface(*m)`
Now: `xxx, err := package.UnmarshalTypeInterface(*m)`

PR #590, opened 2021-01-04
Entity Getters + Increased requirements for EntityInterface

jthurman42:
This creates `Get...` for all fields on all Entities, and updates the EntityInterface to require `GetName` and `GetGUID` which are common across all entity types (there are probably other common fields as well).

Notes:
* This comes out of some merge / usability issues when [getting v0.53.0 into the terraform provider](https://github.com/newrelic/terraform-provider-newrelic/pull/1071), where having `GetName` would be really useful to avoid a lot of type assertion boiler plate
* Not entirely sold on the ergonomics of the getters returning a pointer.
* EntityInterface and EntityOutlineInterface are tied together as we could get either of those interfaces back depending of we Fetch or Search.


templates/typegen/types.go.tmpl
```diff
@@ -42,6 +42,18 @@ type {{.Name}} struct {
   {{    .Name }} {{ .Type }} {{ .Tags }}
   {{- end}}
 }
+{{-  if .GenerateGetters }}
+{{-   range .Fields }}
+// Get{{ .Name }} returns a pointer to the value of {{ .Name }} from {{ $typeName }} or nil
+func (x *{{ $typeName }}) Get{{ .Name}}() *{{ .Type }} {
```
jthurman42, line 48: Like I said, this feels a bit odd to return a pointer.  My reasoning was two fold.  Gotta nil check the receiver and return something (empty value or what not), and decrease creation of a ton of objects in memory for little gain.

Issues of course would be that you could change the value, which could lead to some pretty hard to find bugs...

Alternatives could be to return the value and an error / nil.  Which is arguably the best option now that I'm typing this comment...

.tutone.yml
```diff
@@ -323,18 +313,94 @@ packages:
       - name: EntityTagValueWithMetadata
       - name: EntityTagWithMetadata
       - name: EntityType
+
+      # List of entities
+      - name: Entity
+        interface_methods:
```
zlesnr, line 319: Where is this being used currently?

jthurman42: Alerts: https://github.com/newrelic/newrelic-client-go/pull/590/files/ab2f926234458e74e8f366081573e2e1cbdb9158#diff-c8fb5f06f26ca2ea1554da789ba1914c18e334991add8cbd66482c4568907a17R96

zlesnr: Okay, no worries.  I was going to complain, but this seems like a reasonable step.

zlesnr (approved):

ctrombley (commented):
LGTM.  Wondering if we might consider pattern matchers at some point for setting options across a large number of types.

PR #593, opened 2021-01-05
chore(tools): move dep tracking for tools to tools package

jthurman42:
Moving dep tracking for pipeline tooling to `tools/go.mod` so they aren't hard deps on the main package.  I believe this is causing some issues with importers of the client library, specifically terraform-provider-newrelic.


build/deps.mk
```diff
@@ -3,37 +3,30 @@
...
 VENDOR_CMD   ?= ${GO} mod tidy
 BUILD_DIR    ?= ./bin/
 
-# These should be mirrored in /tools.go to keep versions consistent
-# GOTOOLS      += github.com/client9/misspell/cmd/misspell
+# Go file to track tool deps with go modules
+TOOL_DIR     ?= tools
+TOOL_CONFIG  ?= $(TOOL_DIR)/tools.go
 
+GOTOOLS ?= $(shell cd $(TOOL_DIR) && go list -f '{{ .Imports }}' -tags tools |tr -d '[]')
```
jthurman42, line 14: Neet trick, any imports defined in this package will become tools to install. No more need to `+=` them constantly

zlesnr (approved):

build/deps.mk
zlesnr: Yeah, I like that.

PR #602, opened 2021-01-11
fix(dashboards): DashboardWidgetConfigurationInput needs to be nullable

jthurman42:
NON GENERATED CHANGES

Need to make all fields in `DashboardWidgetConfigurationInput` nullable, otherwise we can only create `markdown` widgets.

This needs to be worked into the code gen, but need this for TF stat.

### DIFF
--- a/pkg/dashboards/types.go
+++ b/pkg/dashboards/types.go
@@ -166,7 +166,7 @@ type DashboardLineWidgetConfigurationInput struct {
 // DashboardMarkdownWidgetConfigurationInput - Configuration for visualization type 'viz.markdown'
 type DashboardMarkdownWidgetConfigurationInput struct {
 	// Markdown content of the widget
-	Text string `json:"text"`
+	Text string `json:"text,omitempty"`
 }
 
 // DashboardPageInput - Page input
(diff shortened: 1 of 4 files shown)


sanderblue (approved):

sanderblue (approved):

pkg/dashboards/types.go
```diff
@@ -212,19 +212,19 @@ type DashboardUpdateResult struct {
 // DashboardWidgetConfigurationInput - Typed configuration for known visualizations. At most one may be populated.
 type DashboardWidgetConfigurationInput struct {
```
jthurman42, line 213: @zlesnr This is a good one to think about how to handle in tutone.  I.e. we want all the nullable structs within this struct to be pointers.

pkg/dashboards/types.go
```diff
@@ -166,7 +166,7 @@ type DashboardLineWidgetConfigurationInput struct {
 // DashboardMarkdownWidgetConfigurationInput - Configuration for visualization type 'viz.markdown'
 type DashboardMarkdownWidgetConfigurationInput struct {
 	// Markdown content of the widget
-	Text string `json:"text"`
+	Text string `json:"text,omitempty"`
```
jthurman42, line 169: Should go back and test if this is required now that `Markdown` below would be `nil`

sanderblue (approved):

pkg/dashboards/types.go
sanderblue: I could've sworn we account for nullable fields with pointers somewhere, but something could be slightly different with introspection on this.

PR #603, opened 2021-01-11
fix(http): update error message for 401 status code to account for missing API key AND invalid API key

sanderblue:
Resolves: #605

Note: We'll need to release the Terraform provider with this update to satisfy the parent ticket referenced in #605

### DIFF
--- a/internal/http/graphql.go
+++ b/internal/http/graphql.go
@@ -2,6 +2,7 @@ package http
 
 import (
 	"encoding/json"
+	"net/http"
 	"strings"
 )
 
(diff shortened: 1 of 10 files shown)


internal/http/graphql.go
```diff
@@ -20,6 +21,7 @@ type GraphQLError struct {
 	Path       []string `json:"path,omitempty"`
 	Extensions struct {
 		ErrorClass string `json:"errorClass,omitempty"`
+		ErrorCode  string `json:"error_code,omitempty"`
```
sanderblue, line 24: Example NerdGraph response that was returned during testing.

```json
{
	"errors": [{
		"extensions": {
			"error_code": "BAD_API_KEY"
		},
		"message": "Bad API Key or no API Key provided"
	}]
}
```

ctrombley (approved):
:shipit:

Issue #612, opened 2021-01-13
Getting JSON Unmarshal Errors from TF Plan using NR Provider


MAN98:
### Description
I am getting some weird errors when I run my CI build which executes a Terraform Plan against my code that uses v 2.7.1 of the NR Provider. This started happening today only and I did not have any errors with my CI build before this.

### Go Version
Not sure what the NR Provider uses for Go version

### Current behavior
TF Plan fails with Go related errors

### Expected behavior
No errors with JSON marshaling

### Steps To Reproduce
1. tf plan on terraform code using NR Provider

### Debug Output (if applicable)

Error: json: cannot unmarshal number into Go struct field entitySearchResponse.data.actor of type nrtime.Minutes
--
1349 |  
1350 | on modules/synth_script_api/main.tf line 48, in data "newrelic_entity" "entity":
1351 | 48: data "newrelic_entity" "entity" {
1352 |  
1353 |  
1354 |  
1355 | Error: json: cannot unmarshal number into Go struct field entitySearchResponse.data.actor of type nrtime.Minutes
1356 |  
1357 | on modules/synth_script_api/main.tf line 48, in data "newrelic_entity" "entity":
1358 | 48: data "newrelic_entity" "entity" {
1359 |  
1360 |  
1361 |  
1362 | Error: Service Exception: EntitySearchService.API.Error Server Error (504)
1363 |  
1364 | on modules/synth_script_api/main.tf line 48, in data "newrelic_entity" "entity":
1365 | 48: data "newrelic_entity" "entity" {

MAN98:
Seems to happen even if I use the later NR Provider version of 2.14.1

zlesnr:
Oh ouch.  Maybe the API changed underneath us.  I'll check on that.

zlesnr:
Is this happening consistently?  Can I see more detail for the `newrelic_entity` resource please?

zlesnr:
Also, can you confirm the version.  I just checked, and the `v2.7.1` of the provider wouldn't have included the `nrtime.Minutes` type yet.

MAN98:
@zlesnr - Yes it is happening consistently since today. I just re-ran our CI and still getting the same errors. Here's how we are using the newrelic_entity resource in our code:
`#entity tags (replaces synthetic labels)
data "newrelic_entity" "entity" {
  name = var.name
}

resource "newrelic_entity_tags" "tags" {
  guid = data.newrelic_entity.entity.guid

[...]

PR #615, opened 2021-01-14
tests(entities): add search that includes a synthetics entity

zlesnr:


zlesnr:
Relates to https://github.com/newrelic/newrelic-client-go/issues/612

ctrombley (approved):

.tutone.yml
```diff
@@ -592,6 +592,8 @@ packages:
         create_as: string
       - name: Seconds
         create_as: string
```
jthurman42, line 594: Do we know that `Seconds` as string won't bite us either?

zlesnr: We do not.  Minutes is a scalar, but comes back in the JSON as an int.

zlesnr: Seconds also is a scalar.  I'll look for some usage of it.

zlesnr: https://github.com/newrelic/newrelic-client-go/issues/617

PR #617, opened 2021-01-14
fix(entities): unmarshal Seconds as float64

zlesnr:


sanderblue (approved):

sanderblue (approved):

.tutone.yml
```diff
@@ -587,13 +587,11 @@ packages:
...
       - name: Minutes
-        create_as: string
+        create_as: int
       - name: NaiveDateTime
         create_as: string
       - name: Seconds
-        create_as: string
-      - name: Minutes
-        create_as: int
+        create_as: float64
```
jthurman42, line 594: Should probably find a test that verifies `Seconds` comes back as a float, since it's a Scalar in GQL and could be anything...

zlesnr: I agree.  Some return `0` which I want to confirm is a valid float.

zlesnr:
We're not currently querying deep enough to hit this in our test environment.  Closing until later.

PR #631, opened 2021-01-20
feat(alerts): Add muting rule schedule fields

akane0915:
https://newrelic.atlassian.net/browse/AINTER-7589

Added the following fields and types to Alerts Muting Rules:

```
schedule: {
    endRepeat:

    endTime:

    repeat:

    repeatCount:

    startTime:

    timeZone:

    weeklyRepeatDays:
}
```

Goals: 
* Fields and types available via Go Client should match what’s currently exposed via NerdGraph.
* Add unit tests
* Update integration tests


ahornerr (approved):
Looks great, nice job!

pkg/alerts/muting_rules.go
```diff
@@ -35,23 +42,113 @@ type MutingRuleCondition struct {
...
+	WEDNESDAY: "WEDNESDAY",
+	THURSDAY:  "THURSDAY",
+	FRIDAY:    "FRIDAY",
+	SATURDAY:  "SATURDAY",
+	SUNDAY:    "SUNDAY",
+}
+
+// NaiveDateTime wraps `time.Time` to remove the time zone offset when JSON marshaling.
+// NaiveDateTime is used for MutingRuleScheduleCreateInput and MutingRuleScheduleUpdateInput fields StartTime, EndTime, and EndRepeat.
+type NaiveDateTime struct {
```
zlesnr, line 86: Do we also need an `UnmarsalJSON` method for this type?

Also, what do you think about moving this type to the `nrtime` package?  If its only used in nerdgraph by the alerts endpoint, here seems fine, but if its used by other types that we might want to reference in the future, perhaps the `nrtime` package would be a better spot.

akane0915: @zlesnr We shouldn't need an `UnmarsalJSON` for this type because the API returns a normal DateTime.  The muting rules fields that use NaiveDateTime type for the incoming payload are a bit strange because the API actually returns a normal DateTime (with an offset) in the response.  So to be accurate to API, I think we should expose that to users (no unmarshaling).

In response to moving it to the `nrtime` package, I think that's fine.  I know nerdgraph created the type specifically for our use case so I don't know if it's used elsewhere, but I don't think it would hurt to move it to the `nrtime` package in case it is used elsewhere in the future.

zlesnr: Okay, that sounds reasonable to leave it in place.

zlesnr (approved):
Nice work, this looks good to me.  Thank you for including the tests as well.

PR #636, opened 2021-01-22
feat(dashboards): add linked entities to nr1 dashboard widgets

sanderblue:
Related to: [newrelic/terraform-provider-newrelic #1128](https://github.com/newrelic/terraform-provider-newrelic/issues/1128)

### DIFF
--- a/pkg/dashboards/dashboard.go
+++ b/pkg/dashboards/dashboard.go
@@ -96,11 +96,21 @@ const getDashboardEntityQuery = `query ($guid: EntityGuid!) {
             }
             layout { column height row width }
             title
-            visualization { id }
+						visualization { id }
+						id
 						linkedEntities {
-							__typename
-							guid
-						}
+              __typename
+              guid
+              name
+              accountId
+              tags {
+                key
+                values
+              }
+              ... on DashboardEntityOutline {
+                dashboardParentGuid
+              }
+            }
           }
         }
         permissions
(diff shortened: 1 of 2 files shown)


pkg/dashboards/dashboard.go
```diff
@@ -97,7 +97,10 @@ const getDashboardEntityQuery = `query ($guid: EntityGuid!) {
             layout { column height row width }
             title
             visualization { id }
-            id
+						linkedEntities {
+							__typename
+							guid
```
sanderblue, line 102: Should we include more fields here?

ctrombley (dismissed):

pkg/dashboards/dashboard.go
jthurman42: Need to keep `id`, which appears to have been dropped.  Can you convert to spaces for consistency within the string?

`GUID` is the only required input field for adding the link to a widget, `name` and `accountId` would be useful for validation / human consumption.  The data set would be rather small on fetch, so adding more fields should have minimal impact if any.  Maybe something like this:

```
            linkedEntities {
              __typename
              guid
              name
              accountId
              tags {
                key
                values
              }
              ... on DashboardEntityOutline {
                dashboardParentGuid
              }
            }
```

Dashboard entities are the only supported type today, might be helpful to have the parent GUID, might never be used.

jthurman42 (dismissed):
Excellent! Ship or enhance.

pkg/dashboards/dashboard.go
sanderblue: Oh crap, I didn't even realize I removed that. Good catch.

sanderblue: Thanks for the suggestions. Updated

ctrombley (approved):
:shipit:

PR #652, opened 2021-01-28
(chore) remove omitEmpty from MutingRuleScheduleUpdateInput and MutingRuleUpdateInput.Schedule

jaymcgrath:
This removes omitEmpty from all of the fields in the mutingRuleScheduleUpdateInput struct.
These fields will be managed as empty by Terraform even omitted from a terraform config.

Removing the omitEmpty tag allows these fields to be nulled if they were set and then removed from a user's terraform config.

Removing this tag from MutingRuleUpdate input allows the entire schedule object to be nulled out if that block is removed from a terraform config.

### DIFF
--- a/pkg/alerts/muting_rules.go
+++ b/pkg/alerts/muting_rules.go
@@ -121,13 +121,13 @@ type MutingRuleScheduleCreateInput struct {
 
 // MutingRuleScheduleUpdateInput is the time window when the MutingRule should actively mute violations for Update
 type MutingRuleScheduleUpdateInput struct {
-	StartTime        *NaiveDateTime            `json:"startTime,omitempty"`
-	EndTime          *NaiveDateTime            `json:"endTime,omitempty"`
-	TimeZone         *string                   `json:"timeZone,omitempty"`
-	Repeat           *MutingRuleScheduleRepeat `json:"repeat,omitempty"`
-	EndRepeat        *NaiveDateTime            `json:"endRepeat,omitempty"`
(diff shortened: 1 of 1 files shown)


jthurman42:
@jaymcgrath the commit message is missing a `:`, can you please update it to:

`refactor(alerts): remove omitEmpty from MutingRuleScheduleUpdateInput`

pkg/alerts/muting_rules.go
```diff
@@ -121,13 +121,13 @@ type MutingRuleScheduleCreateInput struct {
 
 // MutingRuleScheduleUpdateInput is the time window when the MutingRule should actively mute violations for Update
 type MutingRuleScheduleUpdateInput struct {
```
jthurman42, line 123: I'm surprised this file isn't code generated, but in this case it is fortunate for now.  We'll need to solve this on the generation side as well so we can move away from the hand-crafted structs.

jaymcgrath:
> @jaymcgrath the commit message is missing a `:`, can you please update it to:
> 
> `refactor(alerts): remove omitEmpty from MutingRuleScheduleUpdateInput`

Updated, thanks!

jaymcgrath:
@jthurman42 looks like a snyk failure is causing this to fail CI checks

jthurman42:
Synk can't authenticate / run from a third-party fork for security reasons.  No big deal though.  Rebase and we can ignore that check.

jaymcgrath:
@jthurman42 - Thanks! Just force pushed a rebase :)

sanderblue (approved):

PR #655, opened 2021-01-28
refactor(workloads): Generate workload code, deprecate old functions

jthurman42:
Generate `Workloads` code, and deprecate manually generated code.


ctrombley (approved):
🥇

Issue #663, opened 2021-02-11
Correct Insights EU Endpoint for dependant PCF nozzle


stephencrowley:
### Description
Correct Insights EU Endpoint for dependent PCF nozzle

### Current behavior
EU API endpoint is incorrect

### Expected behavior
Corrected EU API endpoint

Please update
https://github.com/newrelic/newrelic-client-go/blob/master/pkg/region/region_constants.go#L36

https://insights-collector.eu.newrelic.com/v1
To 
https://insights-collector.eu01.nr-data.net/v1

As per 
https://docs.newrelic.com/docs/using-new-relic/welcome-new-relic/get-started/our-eu-us-region-data-centers#endpoints

PR #667, opened 2021-02-12
fix(region_constants): corrected insightsBaseURL for EU

harrykimpel:


harrykimpel:
closes #663

zlesnr:
Thank you for the PR.  The commit message here needs to be formatted according to this doc.  https://github.com/newrelic/newrelic-client-go/blob/master/CONTRIBUTING.md#commit-messages

harrykimpel:
@zlesnr updated the commit message

zlesnr:
Thank you for the update @harrykimpel.  The first commit is still in place, so the lint will fail.  Can you squash those together so that all of the commits match the regex please?  Also a rebase on master while you are there since we're a little behind on this branch now.

PR #674, opened 2021-02-16
feat(alerts): adding id to alertsMutingRulesQuery

dannybrody:
currently the `alertsMutingRulesQuery` does not include the rule ids, which is needed for the following methods:
- GetMutingRule
- UpdateMutingRule
- DeleteMutingRule

### DIFF
--- a/pkg/alerts/muting_rules.go
+++ b/pkg/alerts/muting_rules.go
@@ -271,6 +271,7 @@ const (
 			account(id: $accountID) {
 				alerts {
 					mutingRules {
+						id
 						name
 						description
 						enabled


dannybrody:
I also went through the process of signing the CLA

zlesnr:
Thank you for the PR.  For the lint check to pass, the commit needs to formatted as noted in https://github.com/newrelic/newrelic-client-go/blob/master/CONTRIBUTING.md#commit-messages .

dannybrody:
ive tried adding a new commit with the appropriate regex for the commit message, still appears to be failing.  also updated the original title to "feat(alerts): adding id to alertsMutingRulesQuery" not sure if i should just trash this whole branch and start over.  Any guidance?

PR #681, opened 2021-02-18
fix(dashboards): Return rawConfiguration on get

jthurman42:
There are multiple visualizations, plus config options for the `Configuration` versions that are only available in the `rawConfiguration` field, which we do not currently fetch.

This adds the field to the fetch, and a test to ensure that it's returned.


jthurman42:
Blocks newrelic/terraform-provider-newrelic#1208

sanderblue (approved):

PR #686, opened 2021-02-18
fix(dashboards): return an error.NotFound instead of nil

jthurman42:
Should of done this the first time, didn't, now have to back track.

Return that we didn't find it as an error.


ctrombley (approved):
🌮

PR #688, opened 2021-02-20
remove over-strict cast in http.NewClient

novas0x2a:
both Config and Client use RoundTrippers, but this typecheck enforces
that the RoundTripper is actually an http.Transport. Eliminating the
typecheck permits the Transport to be a mock.


novas0x2a:
Hm, I'm not sure I want to provide my home address for the cla. However, this bugfix is simple enough for someone to cherry-pick it themselves.

zlesnr:
Thank you for the PR.  If you have ambitions to send further PRs to implement, say a mock HTTPTransport, the CLA is nice.  I don't remember the home address requirement, and I probably hesitated there too, but I think it's just to know for legal reasons.  In case you want to sign it, we'll need the commit message to match the style here: https://github.com/newrelic/newrelic-client-go/blob/master/CONTRIBUTING.md#commit-messages

ctrombley:
Cherry-picked in #744. Thanks @novas0x2a !

PR #714, opened 2021-04-26
fix(graphql): retry on server errors

ctrombley:


zlesnr (approved):

Julien4218 (approved):
I'm approving, but I have no idea that this will cover the particular issue we're trying to fix.

Issue #721, opened 2021-05-12
Use GetWithContext in apm pkg


Tarliton:
### Feature Description
Hello, 
Could we get a version of `apm` pkg functions that use `context.Context`?

For example, I have a instrumented `http.RoundTripper` that I have configured with `ConfigHTTPTransport`, but when I call functions from `apm` pkg I am not seeing those requests in New Relic Transaction.

I believe the reason is that we are not calling it with the context for the request.

If that is not the reason, please could you guys point me in the right direction?

### Describe Alternatives
Maybe something like [this](https://github.com/newrelic/newrelic-client-go/blob/2932f5d6275d9017fd1ce5764d2de258575dd187/pkg/nerdgraph/nerdgraph.go#L42-L50).

### Additional context
I would like to see in New Relic Transactions what endpoints I am calling when I use this library.

Thank you

zlesnr:
I believe this could be done in a pretty straight forward manner.  Similar to how we do in the `alerts` package, as well is elsewhere probably.

https://github.com/newrelic/newrelic-client-go/blob/main/pkg/alerts/channels.go#L77-L102

Where the non-context-aware method calls the context-aware method to keep compatibility with older implementations, while at the same time providing new context methods to allow what you are after and keeping the logic the same.

Based on what I see here, it doesn't look like too many methods would need to be modified for this:
https://pkg.go.dev/github.com/newrelic/newrelic-client-go@v0.58.5/pkg/apm

Is this a contribution you'd like to make @Tarliton?

And to confirm, are you seeing transactions for the other packages that do implement the context handling?

Tarliton:
Hello @zlesnr, 

Thank you for your reply. Yes, I would like to contribute.

I will also check if we can see transactions when we use methods with context handling (I do not use any of them now, hence I can't answer at the moment).


To confirm, I will create a new `*WithContext` function for each APM struct function. In this case it will be a new function for all the following functions:
```go
func (a *APM) CreateDeployment(applicationID int, deployment Deployment) (*Deployment, error)
func (a *APM) CreateLabel(label Label) (*Label, error)
func (a *APM) DeleteApplication(applicationID int) (*Application, error)
func (a *APM) DeleteDeployment(applicationID int, deploymentID int) (*Deployment, error)
func (a *APM) DeleteLabel(key string) (*Label, error)
func (a *APM) GetApplication(applicationID int) (*Application, error)
[...]

PR #727, opened 2021-05-20
Allow passing context to apm methods

Tarliton:
Hello

As discussed [here](https://github.com/newrelic/newrelic-client-go/issues/721), this PR adds a version of all APM methods that accepts a context parameter.

I did not add a version for applicationsREST internal functions because, well, they are internal.

Thanks!

### DIFF
--- a/pkg/apm/application_instances.go
+++ b/pkg/apm/application_instances.go
@@ -1,6 +1,9 @@
 package apm
 
-import "fmt"
+import (
+	"context"
+	"fmt"
+)
 
 // ApplicationInstanceLinks represents all the links for a New Relic application instance.
 type ApplicationInstanceLinks struct {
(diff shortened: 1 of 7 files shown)


zlesnr:
This looks good to me.  Thank you for putting this together.  To get the tests to pass, `make lint` is noticing that two methods aren't used any more.  If you remove them, this should be a bit closer.

Tarliton:
> This looks good to me. Thank you for putting this together. To get the tests to pass, `make lint` is noticing that two methods aren't used any more. If you remove them, this should be a bit closer.

Oh, okay. I'll remove them.

ctrombley (approved):

ctrombley:
Thanks @Tarliton !

PR #771, opened 2021-07-09
Fix tutone generation

jthurman42:
When resolving an issue with billboard input, a type was skipped that is needed.  Update the tutone config to generate the old type, and skip the overridden type.  [Reference](https://github.com/newrelic/newrelic-client-go/commit/c84c84690fd363ecf0b7fa6074359ebc9af4c803#diff-c8fb5f06f26ca2ea1554da789ba1914c18e334991add8cbd66482c4568907a17R208)

### DIFF
--- a/.tutone.yml
+++ b/.tutone.yml
@@ -205,7 +205,6 @@ packages:
         field_type_override: "*DashboardBarWidgetConfigurationInput"
       - name: DashboardBillboardWidgetConfigurationInput
         field_type_override: "*DashboardBillboardWidgetConfigurationInput"
-        skip_type_create: true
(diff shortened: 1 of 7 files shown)


sanderblue (dismissed):
🐑 it

jthurman42:
@mbazhlekova This changes the error type returned, but also updates Tutone to generate what you had done previously.  This could be improved though, to use the newrelic-client-go errors package and add some better wrapping functions, and allow for type assertion on them.

mbazhlekova (approved):

mbazhlekova (approved):

PR #773, opened 2021-07-09
chore(changelog): Update CHANGELOG for v0.61.0

zlesnr:


abrunner94 (approved):

abrunner94:
Got to hold off on this one. We need to change the mutation from `createRecipeEvent(...)` to `installationCreateRecipeEvent(...)`. Totally missed that, but the installation part is the namespace in NerdGraph.

PR #780, opened 2021-07-22
fix(logging): export LogrusLogger for use in other projects

ctrombley:
This will be used by https://github.com/newrelic/newrelic-cli/pull/1005 to pass in the logrus instance already defined in the CLI.


sanderblue (approved):
🐑 it

Julien4218 (approved):

PR #785, opened 2021-07-27
fix: retire usages of deprecated error schema

### LINKED ISSUE #765: Error handling updates needed (deadline of 07/31)
Nerd Graph's error structure is changing slightly for stitched services. Check the announcements in `#nerdgraph-announcements` for specific details on what's changing and how clients will need to adapt.

Since Nerd Graph is currently breaking spec to ensure backwards compability we'll need to keep this transition window relatively short. We've set a deadline of one month (07/31) for folks to transition.

Someone with more familiarity should probably audit the error handling code to build confidence, but I think I've spotted at least a couple places that will need an update. It'd be best if people with more experience did a quick check as well. More details are in the blog posts posted in the slack channel above.

https://github.com/newrelic/newrelic-client-go/blob/2932f5d6275d9017fd1ce5764d2de258575dd187/pkg/alerts/nrql_conditions.go

This error handling code will break because the downstreamResponse key is going away. Instead, this code should iterate through the errors list and check the "message" value of each error. In other words, drilling down into "downstream response" is not necessary. Error objects formerly in `downstreamResponse` will now be presented directly, with things like error extensions intact. No need to "drill down".

We're not Go experts on Unified API, but maybe the fix could be as simple as this?

```
	for _, err := range r.Errors {
    if strings.Contains(err.Message, "Not Found") {
				return true
			}
	}
```

Note that this would rely on an update to `/internal/http/graphql.go`.

https://github.com/newrelic/newrelic-client-go/blob/2932f5d6275d9017fd1ce5764d2de258575dd187/pkg/alerts/policies.go

This is the same situation as the above. A fix might look something like?

```
for _, err := range r.Errors {
  if err.Message == "Not Found" &&
    err.Extensions.Code == "BAD_USER_INPUT" {
    return true
  }
}
```

Note that this would rely on an update to `/internal/http/graphql.go`.

https://github.com/newrelic/newrelic-client-go/blob/2932f5d6275d9017fd1ce5764d2de258575dd187/internal/http/graphql.go

Errors from downstream services are being merged in with any other errors Nerd Graph sends and the downstreamResponse key will be removed.

This means the list of errors in a Nerd Graph response should be iterated through and this code should check the `extensions` property of each error instead of drilling down into the "downstreamResponse". The "path" can help you identify if a specific error maps to a specific field.

If there are lingering questions after the blog posts, please come find us in our help channel! Unified API wants to make this as pain-free as possible.

ctrombley:
Resolves #765 .

### DIFF
--- a/internal/http/graphql.go
+++ b/internal/http/graphql.go
@@ -1,7 +1,6 @@
 package http
 
 import (
-	"encoding/json"
 	"net/http"
 	"strings"
 )
@@ -23,19 +22,6 @@ type GraphQLError struct {
 		ErrorClass string `json:"errorClass,omitempty"`
 		ErrorCode  string `json:"error_code,omitempty"`
 	} `json:"extensions,omitempty"`
-	DownstreamResponse []GraphQLDownstreamResponse `json:"downstreamResponse,omitempty"`
-}
-
-// GraphQLDownstreamResponse represents an error's downstream response.
-type GraphQLDownstreamResponse struct {
-	Extensions struct {
-		Code             string `json:"code,omitempty"`
-		ValidationErrors []struct {
-			Name   string `json:"name,omitempty"`
-			Reason string `json:"reason,omitempty"`
-		} `json:"validationErrors,omitempty"`
-	} `json:"extensions,omitempty"`
-	Message string `json:"message,omitempty"`
 }
 
 // GraphQLErrorResponse represents a default error response body.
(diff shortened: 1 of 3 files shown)


Julien4218 (approved):

mattbaker (approved):
Awesome, thanks so much @ctrombley 🙏 

We could check with their team, but it looks right to me based on my knowledge of the Alerts API.

PR #803, opened 2021-08-26
feat: Provide additional context in GraphQL errors for Alerts operations

ahornerr:
Alerts via GraphQL operations (specifically condition create/update) supply additional context via `validationErrors` that is not exposed currently.

This change enables enhanced alert condition create/update errors to the Terraform provider.

By combining `internal/http/graphql` and the `nrqlConditionErrorResponse` into a new `alerts.GraphQLErrorResponse` exposing this additional context can be easily accomplished.

`http.GraphQLErrorResponse` was not modified as it is the default NerdGraph error response type, and this additional context is specific to Alerts.

In `internal/http/client.go` the `errorValue` variable is returned as-is, as opposed to converting to a new errorString via `errors.New(errorValue.Error())`. This is necessary to expose the error context downstream, and allows type casting of the returned error. This should not have any negative impact as the `ErrorResponse` type already implements `error.Error()`.


akane0915:
Note: This PR is followed by this [Terraform Provider PR.](https://github.com/newrelic/terraform-provider-newrelic/pull/1426)

PR #809, opened 2021-09-15
feat(servicelevel): update code gen strategy

ctrombley:


sanderblue (dismissed):

asanuy (commented):
Good job @ctrombley, I really appreciate that you took the time to take a look at this! :)

Just a few comments regarding some references to the `slug` that still remain in the code.

I can see that the mutations still hold references to the field, for instance:

https://github.com/newrelic/newrelic-client-go/blob/main/pkg/servicelevel/servicelevel_api.go#L108

Checking the configuration in the tutone library, I seems that it would be possible to exclude fields in mutations using the [excludeFields](https://github.com/newrelic/tutone/blob/main/internal/config/config.go#L118) property. Do you think it would work?

Besides that, I can see that in the `pkg/entities` there are also some references to the field. I guess we could apply similar exceptions as in the `servicelevel` package?

For the records, these are the references in the `pkg/entities`:

https://github.com/newrelic/newrelic-client-go/blob/main/pkg/entities/types.go#L6602
https://github.com/newrelic/newrelic-client-go/blob/main/pkg/entities/entities_api.go#L344
https://github.com/newrelic/newrelic-client-go/blob/main/pkg/entities/entities_api.go#L1141

asanuy (dismissed):
Brilliant, looking sharp now. Thanks for taking care of it! 🍏

asanuy (approved):

jthurman42 (approved):
:shipit:

pkg/servicelevel/types.go
```diff
@@ -251,6 +246,9 @@ type UserReference struct {
 	Name string `json:"name,omitempty"`
 }
 
+// EntityGUID - An encoded Entity GUID
+type EntityGUID string
```
jthurman42, line 250: We should think about moving this out to a very basic module that has no dependancies to solve these circular import issues for common types and not re-defining the type.

PR #810, opened 2021-09-17
refactor: move EntityGUID to a common package

ctrombley:
A [prior PR](#809) introduced a second instance of the EntityGUID type to avoid a circular dependency between the `entities` and `servicelevel` packages.  This PR attempts to refactor the `EntityGUID` type to a common package, but it introduces lots of breaking changes that will need to be addressed in downstream projects if it were to be merged and released.


jthurman42 (approved):
:shipit: 

Minor breaking changes in terraform-provider-newrelic and newrelic-cli

PR #814, opened 2021-09-23
fix(servicelevel): Update code gen strategy

asanuy:
- Add query to fetch SLIs through the entity API
- Exclude resultQueries from responses
- Define some overrides needed
- Refactor some types owned by other packages


pkg/servicelevel/servicelevel.go
```diff
@@ -18,3 +18,18 @@ func New(config config.Config) Servicelevel {
 		logger: config.GetLogger(),
 	}
 }
+
+type EntityInterface struct {
```
asanuy, line 22: Since the [entity query](https://github.com/newrelic/newrelic-client-go/blob/81a8dfcfbfe1c04f0557d88339b14b3da8534734/.tutone.yml#L340-L341) defined in the `entities` package has a field of depth 2 that is insufficient to fetch the necessary data for service levels, I had to define a [new query](https://github.com/newrelic/newrelic-client-go/blob/81a8dfcfbfe1c04f0557d88339b14b3da8534734/.tutone.yml#L845) in the `servicelevel` package itself.

While the code generated is almost OK, the Actor type generated expects an [EntityInterface](https://github.com/newrelic/newrelic-client-go/blob/81a8dfcfbfe1c04f0557d88339b14b3da8534734/pkg/servicelevel/types.go#L26) while the object generated is called just [Entity](https://github.com/newrelic/newrelic-client-go/blob/81a8dfcfbfe1c04f0557d88339b14b3da8534734/pkg/servicelevel/types.go#L32). I played with tutone and tried unsuccessfully to overcome this problem without generating all the objects that implement the interface, which are unnecessary.

So I ended up doing the trick highlighted here that seems to work. Happy to hear your thoughts or any suggestion you think might work. Thanks in advance!

sanderblue (approved):

PR #839, opened 2021-11-22
fix(http): Better handling of Deprecation Errors for NerdGraph requests

jthurman42:
Previously, any error condition coming from NerdGraph caused the request to fail.  This isn't great as field `deprecations` are now being returned as errors, with the result set.  Example Entities query:

```
entities {
  relationships {...}
}
```
Result:
```
This field is deprecated! Please use `relatedEntities` instead.
```

This change implements a custom error method for GraphQL Queries (NerdGraph really) that checks to see if it's a deprecation error and continues anyway.  So now there is a case where you get back a valid response body and an error.

*THIS IS LIKELY TO CAUSE SOME DOWNSTREAM ISSUES* but not sure yet... The updated method to check if an error is valid or not is a bit clumsy, and consists of:

```go
  resp, err := client.GetEntities(guids)

  if e, ok := err.(*http.GraphQLErrorResponse); ok {
    if !e.IsDeprecated() {
      return e
    }
  }
```

### DIFF
--- a/internal/http/client.go
+++ b/internal/http/client.go
@@ -364,8 +364,11 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 		return nil, nrErrors.NewNotFound("resource not found")
 	}
 
-	if errorValue.Error() != "" {
-		return nil, errorValue
+	// Ignore deprecation errors
+	if !errorValue.IsDeprecated() {
+		if errorValue.Error() != "" {
+			return nil, errorValue
+		}
 	}
 
 	if req.value == nil {
--- a/internal/http/client_test.go
+++ b/internal/http/client_test.go
@@ -118,6 +118,10 @@ func (c *CustomErrorResponse) IsRetryableError() bool {
 	return false
 }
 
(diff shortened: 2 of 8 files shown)


.tutone.yml
```diff
@@ -336,14 +336,17 @@ packages:
...
               - slug
           - name: entity
             max_query_field_depth: 2
             exclude_fields:
+              - relationships
               - slug
           - name: entitySearch
             max_query_field_depth: 2
             exclude_fields:
+              - relationships
```
sanderblue, line 349: This seems to just be removing this from the query. Should this also be removing the field(s) on the associated structs?

jthurman42: Oh interesting... I'm not sure why tutone is still adding it to the structs on generation.  Removing from the query resolves the issue, and I guess would not be an API change if we left the field, they would just always be empty 🤷🏻‍♂️

jthurman42:
Pulling back to Draft as there is a better way to do this.

mbazhlekova (approved):

PR #843, opened 2021-12-28
Add license key authorization for the Event API.

fryckbos:
Adding License Key Authorizer for the Event API. To keep backwards compatibility the Insight Insert Key is prioritised over the License Key.

The reason for adding this authorizer is to make it easy in the New Relic CLI to create an event without requiring an Insights Insert key.


sanderblue (approved):
Nice! Thanks for the contribution and including an integration test 🎉

PR #847, opened 2022-01-19
feat(installevents): implement metadata field as arbitrary key-value store on recipe event 

sanderblue:
**Note:** The `metadata` field in NerdGraph is an arbitrary key:value object represented as a GraphQL custom scalar type named `InstallationRawMetadata`. Since Tutone translates scalars as strings, we perform an override in our Tutone config to ensure the proper Go type is generated for this field, which is `map[string]interface{}` 😎

### DIFF
--- a/.tutone.yml
+++ b/.tutone.yml
@@ -631,6 +631,9 @@ packages:
       - name: ID
         field_type_override: string
         skip_type_create: true
+      - name: InstallationRawMetadata
+        field_type_override: map[string]interface{}
+        skip_type_create: true
     mutations:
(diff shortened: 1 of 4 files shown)


Julien4218 (approved):

abrunner94 (approved):

sanderblue:
Failing integration tests are unrelated to this PR. 

Confirmed the `installevents` [integration tests](https://github.com/newrelic/newrelic-client-go/runs/4874897062?check_suite_focus=true#step:6:396) are still passing. 

Merging since this failing tests are unrelated and outside the scope of this PR.

PR #849, opened 2022-02-09
feat(entities): add new entity types

sanderblue:
**NOTE:** This PR required some tricky code generation for this to work. This particular set of new types contains a GraphQL `UNION` that we currently don't support in `tutone`, but I was able to make it work with some manual additions such as the `ai` package and other additional necessary manual edits in the `entities/types.go` file.

Additional work needs to be done in Tutone to support the GraphQL `UNION` type.


pkg/ai/types_ai.go
```diff
@@ -0,0 +1,95 @@
+//nolint:revive
+package ai
```
sanderblue, line 2: Note: These types were added generated, but added manually to facilitate fixing unmarshaling error that was happening due to new entity types being introduced in NerdGraph.

mbazhlekova (approved):

PR #850, opened 2022-02-12
chore(nrql_conditions): add test for optional value function for static nrql conditions

akane0915:
https://newrelic.atlassian.net/browse/AINTER-8205

Corresponding TF Provider ticket https://github.com/newrelic/terraform-provider-newrelic/pull/1656

## Goals
**This PR along with the corresponding TF Provider PR requires a MINOR release and should ensure the changelog has an entry noting the NRQL condition `ValueFunction` deprecation.**

This PR adds a test to create and update a static condition with and without a value function.  This work is to support the deprecation of the field `ValueFunction`.  See corresponding TF provider PR.

## Implementation notes: 
- I realized there weren't any code changes needed to support this deprecation, however, I decided to add a test.
- Originally, the goal was to remove `omitempty` from the update struct for `ValueFunction`; however, this was not possible because both static conditions and baseline conditions use the same update struct and value function is not valid for baseline conditions.

### DIFF
--- a/pkg/alerts/nrql_conditions_integration_test.go
+++ b/pkg/alerts/nrql_conditions_integration_test.go
@@ -26,6 +26,7 @@ var (
 	nrqlConditionBaseAggDelay           = 2                                           // needed for setting pointer
 	nrqlConditionBaseAggTimer           = 5                                           // needed for setting pointer
 	nrqlConditionBaseSlideBy            = 30                                          // needed for setting pointer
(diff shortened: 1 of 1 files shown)


jaymcgrath (approved):
This looks good to me. Kudos for a good explanation and thanks for adding an integration test to make sure these changes are backwards compatible

mbazhlekova (approved):

PR #852, opened 2022-02-22
Update example in readme to compile and run with v0.73.0

willcain:
Update example in readme to compile and run with v0.73.0


jthurman42:
@willcain happy to get this merged in if you could update the commit message to be `docs: <same as current message>` so the linter will pass. Info on commit messages is available here: https://github.com/newrelic/newrelic-client-go#commit-messages

willcain:
Thanks, updated.

sanderblue (approved):
Thanks @willcain! 👏

Issue #697, opened 2021-03-11
Add x-account-id header to REST requests


jthurman42:
### Feature Description
Add `X-Account-ID` header on REST requests so the same API key can be used across all account it has access to (not just it's primary account id)

### Describe Alternatives
Create a new client for each account managed for REST resources.

### Additional context
The legacy REST API now supports cross-account resource management via API key with an included `X-Account-ID` header, which is currently not sent.

PR #854, opened 2022-03-01
feat(auth): Add X-Account-ID header if value exists in request context

Fallenstedt:
Closes https://github.com/newrelic/newrelic-client-go/issues/697

This PR adds an X-Account-ID header on REST requests so the same API key can be used across all account it has access to (not just it's primary account id).

To add the header, users should leverage the request's `context` and add an `X-Account-ID` key in the request's context.  This is done with a `contextkeys` package.

This work is meant to address an issue creating channels for child accounts with New Relic's [terraform-provider-newrelic](https://github.com/newrelic/terraform-provider-newrelic/issues/1645).

### DIFF
--- a/internal/http/auth.go
+++ b/internal/http/auth.go
@@ -2,6 +2,7 @@ package http
 
 import (
 	"github.com/newrelic/newrelic-client-go/pkg/config"
+	"github.com/newrelic/newrelic-client-go/pkg/contextkeys"
 )
 
 // RequestAuthorizer is an interface that allows customizatino of how a request is authorized.
@@ -26,6 +27,10 @@ func (a *PersonalAPIKeyCapableV2Authorizer) AuthorizeRequest(r *Request, c *conf
 		r.SetHeader("Api-Key", c.PersonalAPIKey)
 	} else {
 		r.SetHeader("X-Api-Key", c.AdminAPIKey)
+	}
+
+	if xAccountID, ok := contextkeys.GetXAccountID(r.request.Context()); ok == true {
+		r.SetHeader("X-Account-ID", xAccountID)
 	}
 }
 
--- a/internal/http/auth.go
+++ b/internal/http/auth.go
@@ -29,7 +29,7 @@ func (a *PersonalAPIKeyCapableV2Authorizer) AuthorizeRequest(r *Request, c *conf
 		r.SetHeader("X-Api-Key", c.AdminAPIKey)
 	}
 
-	if accountID, ok := contextkeys.GetAccountID(r.request.Context()); ok == true {
+	if accountID, ok := contextkeys.GetAccountID(r.request.Context()); ok {
 		r.SetHeader("X-Account-ID", accountID)
 	}
 }
--- a/internal/http/auth.go
+++ b/internal/http/auth.go
@@ -29,8 +29,8 @@ func (a *PersonalAPIKeyCapableV2Authorizer) AuthorizeRequest(r *Request, c *conf
 		r.SetHeader("X-Api-Key", c.AdminAPIKey)
 	}
 
-	if xAccountID, ok := contextkeys.GetXAccountID(r.request.Context()); ok == true {
-		r.SetHeader("X-Account-ID", xAccountID)
+	if accountID, ok := contextkeys.GetAccountID(r.request.Context()); ok == true {
+		r.SetHeader("X-Account-ID", accountID)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


internal/http/client.go
```diff
@@ -385,6 +385,12 @@ func (c *Client) Do(req *Request) (*http.Response, error) {
 
 func (c *Client) innerDo(req *Request, errorValue ErrorResponse, i int) (*http.Response, []byte, bool, error) {
 	r, err := req.makeRequest()
+
+	xAccountId := req.request.Context().Value("XAccountID")
+	if xAccountId != nil {
+		req.SetHeader("X-Account-ID", xAccountId.(string))
+	}
```
jthurman42, line 392: As this is auth logic, I suggest moving this to the [`PersonalAPIKeyCapableV2Authorizer`](https://github.com/newrelic/newrelic-client-go/blob/main/internal/http/auth.go#L29) that is used by [Alerts](https://github.com/newrelic/newrelic-client-go/blob/main/pkg/alerts/alerts.go#L24)), which is called inside of `req.makeRequest()` above.

internal/http/auth.go
```diff
@@ -27,6 +28,10 @@ func (a *PersonalAPIKeyCapableV2Authorizer) AuthorizeRequest(r *Request, c *conf
 	} else {
 		r.SetHeader("X-Api-Key", c.AdminAPIKey)
 	}
+
+	if accountID, ok := contextkeys.GetAccountID(r.request.Context()); ok == true {
```
jthurman42, line 32: This can end with just ` ; ok {`

pkg/contextkeys/contextkeys.go
```diff
@@ -0,0 +1,41 @@
...
+func SetAccountID(ctx context.Context, value string) context.Context {
+	return keys.setAccountID(ctx, value)
+}
+
+// GetAccountID returns the account ID from the context.
+func GetAccountID(ctx context.Context) (string, bool) {
+	return keys.getAccountIDFromContext(ctx)
+}
+
+func (c contextKeys) getAccountIDFromContext(ctx context.Context) (string, bool) {
```
jthurman42, line 26: This should match the package globals above for consistency (and be exported for other uses) `GetAccountID`

pkg/contextkeys/contextkeys.go
```diff
@@ -0,0 +1,41 @@
...
+	if nil != ctx {
+		accountID, ok := ctx.Value(keys.accountID).(string)
+		return accountID, ok
+	}
+
+	return "", false
+}
+
+func (c contextKeys) setAccountID(ctx context.Context, value string) context.Context {
```
jthurman42, line 35: This should match the package globals above for consistency (and be exported for other uses) `SetAccountID`

pkg/contextkeys/contextkeys.go
```diff
@@ -0,0 +1,41 @@
...
+	if nil == ctx {
+		ctx = context.Background()
+	}
```
jthurman42, line 38: I believe it's possible that the `accountID` struct underlying this could also be nil.  I don't see a test for the underlying `contextKey.Get/Set` functions for that case.

pkg/contextkeys/contextkeys_test.go
```diff
@@ -0,0 +1,42 @@
...
+// +build unit
+
+package contextkeys
+
+import (
+	"context"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
```
jthurman42, line 11: Should add a test for using the contextKey outside of the package (not sure why someone might do that, but it's available)

```
x := contextKey{}
x.SetAccountID(context.Background(), "blah") // This might panic?
```

jthurman42 (changes requested):

pkg/contextkeys/contextkeys.go
Fallenstedt: I feel like accountID struct can't be nil. I get the type error `Cannot convert 'nil' to type 'struct{}'` when I attempt to add the following: 

```
	if nil == c.accountID {
		c.accountID = struct{}{}
	}
```

I think it could only be nil if the type `contextKeys.accountID` was a pointer like this:
```
type contextKeys struct {
	accountID *struct{}
}
```

jthurman42: fair enough.

jthurman42 (approved):
:shipit:

PR #859, opened 2022-03-02
Fixes #851

### LINKED ISSUE #851: incorrect schema mapping in Alerts.ListAlertEvents
### Description

Alerts event response doesn't match with the actual response schema. Here https://github.com/newrelic/newrelic-client-go/blob/2932f5d6275d9017fd1ce5764d2de258575dd187/pkg/alerts/events.go#L62 `alert_events` should be `recent_events`

From the API explorer

image

yesoreyeram:
Fixes #851

### DIFF
--- a/pkg/alerts/events.go
+++ b/pkg/alerts/events.go
@@ -60,5 +60,5 @@ func (a *Alerts) ListAlertEventsWithContext(ctx context.Context, params *ListAle
 }
 
 type alertEventsResponse struct {
-	AlertEvents []*AlertEvent `json:"alert_events,omitempty"`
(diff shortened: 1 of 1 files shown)


sanderblue (approved):
Hi @yesoreyeram, thanks for this fix 👍

kidk:
@yesoreyeram It seems your commit message is not in the right format: https://github.com/newrelic/newrelic-client-go/blob/main/CONTRIBUTING.md#commit-messages

Would it be possible to change it? https://stackoverflow.com/questions/8981194/changing-git-commit-message-after-push-given-that-no-one-pulled-from-remote

PR #866, opened 2022-04-08
feat(errors): handle 402 payment required HTTP response scenario

sanderblue:


mbruzina (approved):
Straightforward enough 👍

PR #875, opened 2022-04-27
feat: Expose EntityGUID on NRQL Conditions.

Fallenstedt:
This PR will expose EntityGUID on NRQL Alert Conditions. 

These EntityGUIDs will be exposed to consumers of New Relic's Terraform Provider. 

https://newrelic.atlassian.net/browse/AINTER-8515

In addition to the unit test update, I ran this locally to confirm I could see my EntityGUIDs on existing NRQL Conditions 

```
package main

import (
	"fmt"
	"github.com/newrelic/newrelic-client-go/newrelic"
)

func main() {

	client, err := newrelic.New(newrelic.ConfigPersonalAPIKey("YOUR_API_KEY_HERE"))
	if err != nil {
		fmt.Printf("error initializing client:", err)
	}

	conditions, err := client.Alerts.ListNrqlConditions(<Your Policy ID Number Here>)
	if err != nil {
		fmt.Printf("error listing conditions", err)
	}
	for _, condition := range conditions {
		fmt.Println(*condition.EntityGUID)
	}

}
```

### DIFF
--- a/pkg/alerts/nrql_conditions.go
+++ b/pkg/alerts/nrql_conditions.go
@@ -3,6 +3,7 @@ package alerts
 import (
 	"context"
 	"fmt"
+	"github.com/newrelic/newrelic-client-go/pkg/common"
 
 	"github.com/newrelic/newrelic-client-go/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


pkg/alerts/nrql_conditions.go
```diff
@@ -350,17 +351,18 @@ type NrqlAlertCondition struct {
...
+	ExpectedGroups      int                `json:"expected_groups,omitempty"`
+	ID                  int                `json:"id,omitempty"`
+	ViolationCloseTimer int                `json:"violation_time_limit_seconds,omitempty"`
+	Name                string             `json:"name,omitempty"`
+	Nrql                NrqlQuery          `json:"nrql,omitempty"`
+	RunbookURL          string             `json:"runbook_url,omitempty"`
+	Terms               []ConditionTerm    `json:"terms,omitempty"`
+	Type                string             `json:"type,omitempty"`
+	ValueFunction       ValueFunctionType  `json:"value_function,omitempty"`
+	EntityGUID          *common.EntityGUID `json:"entity_guid,omitempty"`
```
Fallenstedt, line 365: :memo: This is a pointer to `common.EntityGUID` because the `NrqlCondition` struct is used to create conditions too.

When creating conditions, a user will not have the EntityGUID available, and it should be omitted completely from the JSON object.

Fallenstedt, line 365: :memo: I am unsure if `EntityGUID` is the correct name here. Should we use `GUID` instead?

sanderblue: `EntityGUID` works well here. We try to be as consistent as we can with the "field name to json property name" 🙂

pkg/alerts/nrql_conditions_test.go
```diff
@@ -85,6 +112,8 @@ var (
 			"runbook_url": "https://www.example.com/docs"
 		}
 	}`
+
+	entityGuid = common.EntityGUID("NDAwQzA0rEFST1BTfENPTkRJVElPTnw3MzUzNjL")
```
sanderblue, line 116: Can we namespace this variable name to avoid a naming collision in another test? Maybe `testAlertsEntityGUID`?

sanderblue, line 116: Can we namespace this variable name to avoid a naming collision in another test? Maybe `testAlertsEntityGUID`?

sanderblue:
@Fallenstedt Thanks for the contribution. Just one minor suggestion and I think we'll be good to go.

sanderblue:
@Fallenstedt Thanks for the contribution. Just one minor suggestion and I think we'll be good to go.

pkg/alerts/nrql_conditions_test.go
Fallenstedt: @sanderblue The variable has been renamed to `testNrqlConditionEntityGUID`

PR #879, opened 2022-04-28
feat: Expose EntityGUID on NRQL Conditions when using NerdGraph.

Fallenstedt:
Continuing from https://github.com/newrelic/newrelic-client-go/pull/875

This PR will expose EntityGUIDs on Nrql Conditions that are fetched from NerdGraph. I mistakenly forgot to do this in my previous pull request.


pkg/alerts/nrql_conditions.go
```diff
@@ -823,6 +823,7 @@ const (
       query
     }
     enabled
+	entityGuid
     description
```
Fallenstedt, line 827: :memo: I am not sure why the diff looks this way. My IDE shows this is formatted well. 
Screen Shot 2022-04-28 at 1 22 58 PM

sanderblue (approved):
Thanks!

PR #883, opened 2022-05-04
feat(nrqldroprules): add `source` attribute to drop rules

heat1q:
In order to display drop rules added with the NerdGraph API, the `source` field of the rules needs to be set to `Logging`. (see https://docs.newrelic.com/docs/logs/ui-data/drop-data-drop-filter-rules/#nerdgraph)

This change adds an optional `Source` field to the respective structs, such that the user can change the source of the drop rule and also see them when querying them.

### DIFF
--- a/pkg/nrqldroprules/example_basic_test.go
+++ b/pkg/nrqldroprules/example_basic_test.go
@@ -33,6 +33,7 @@ func Example_basic() {
 			Action:      NRQLDropRulesActionTypes.DROP_DATA,
 			Description: "NRQL drop rule description",
 			NRQL:        "SELECT * FROM Log WHERE container_name = 'noise'",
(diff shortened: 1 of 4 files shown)


kidk:
Thanks @heat1q 

The interesting thing for me is that the `source` field is not enabled publicly, but only for product within New Relic. That doesn't mean it won't work, but I'd like to clear that out before accepting this.

heat1q:
Hi @kidk, do you have any updates regarding this change?

kidk:
Thanks for patience @heat1q I'm still chasing this internally with our logging team.

kidk:
Sorry for delay on this. Seems there's no plan to make this field public, which means we cannot support it in client-go, cli or Terraform.

PR #884, opened 2022-05-05
Add synthetics secure credentials GraphQL API

mbazhlekova:


pkg/synthetics/synthetics.go
```diff
@@ -72,3 +72,17 @@ func New(config config.Config) Synthetics {
 
 	return pkg
 }
+
+func NewV2(config config.Config) Synthetics {
```
sanderblue, line 76: Do we need a `NewV2()` here? It's returning the same instance as the original `New()`. In this case, the `Synthetics` struct will have both the legacy methods that are pointed at the REST API as well as the new methods pointed at NerdGraph, which is probably fine. We should add [`Deprecated` labels ](https://go.dev/blog/godoc) to the old method doc blocks in this PR too, so the the client's godocs get updated and so other developers will see that in the IDE hints, etc.

mbazhlekova: That makes sense. I removed `NewV2()` and added some deprecation labels to the old functions.

sanderblue (approved):

PR #892, opened 2022-05-20
chore(alerts): Remove references to nrql outlier condition type (EOL)

jaymcgrath:
Outlier NRQL conditions have reached end of life and are no longer supported. This PR removes references to this NRQL condition type and its fields.


sanderblue:
Thanks @jaymcgrath!

PR #899, opened 2022-06-02
Retry condition for GraphQL API error: TOO_MANY_REQUESTS 

luiscamaral:
Add retry condition for **GraphQL** `TOO_MANY_REQUESTS` error response on `json`.


internal/http/retry.go
```diff
@@ -65,5 +70,17 @@ func RetryPolicy(ctx context.Context, resp *http.Response, err error) (bool, err
...
 	}
 
+	// Check for json response and description for retries.
+	// i.e.: when receiving TOO_MANY_REQUESTS it is a HTTP 200 code with an
+	// error status that needed to be retried.
+	json, jErr := io.ReadAll(resp.Body)
+	resp.Body = io.NopCloser(bytes.NewBuffer(json))
+	if jErr == nil {
+		errorClass := fastjson.GetBytes(json, "errors", "0", "extensions", "errorClass")
+		if errorClass != nil && bytes.Compare(errorClass, []byte("TOO_MANY_REQUESTS")) == 0 {
```
sanderblue, line 80: 🤔 Thoughts on using `bytes.Equal()` in the check here?

internal/http/retry.go
```diff
@@ -65,5 +70,17 @@ func RetryPolicy(ctx context.Context, resp *http.Response, err error) (bool, err
...
+			return true, fmt.Errorf("TOO_MANY_REQUESTS")
```
sanderblue, line 81: Can we make `TOO_MANY_REQUESTS` a `const` in the `internal/http` package. Seems like a good candidate for a package level constant 😄 

And it would also be awesome to have at least one unit test for this addition.

sanderblue (changes requested):

internal/http/retry.go
luiscamaral: Converted to Error constant in the same package. The same is being used on the condition.

internal/http/retry.go
luiscamaral: As the error was changed to a `const` on the package, it looked cleaner to use `GetString` and not bytes anymore.

sanderblue (approved):

PR #905, opened 2022-06-03
fix: remove deprecated field from service level query

lucasmessinar:


mbazhlekova:
@lucasmessinar Thanks for your contribution! Can you also add `slug` to this list of `exclude_fields`? https://github.com/newrelic/newrelic-client-go/blob/main/.tutone.yml#L951

This will ensure that if we do the automatic type generation, your changes won't be undone.

mbazhlekova (approved):

PR #914, opened 2022-06-23
chore(synthetics): use *bool type to avoid removing false values

mbazhlekova:


pkg/synthetics/synthetics_api_integration_test.go
```diff
@@ -106,23 +110,23 @@ func TestSyntheticsSimpleBrowserMonitor_Basic(t *testing.T) {
...
+		Name:   monitorName + "-updated",
 		Period: SyntheticsMonitorPeriod(SyntheticsMonitorPeriodTypes.EVERY_5_MINUTES),
 		Status: SyntheticsMonitorStatus(SyntheticsMonitorStatusTypes.ENABLED),
 		Tags: []SyntheticsTag{
 			{
-				Key: "Name",
+				Key: "pineapple",
 				Values: []string{
-					"synthetics",
+					"pizza",
```
kidk, line 129: 😆

pkg/synthetics/synthetics_api_integration_test.go
```diff
@@ -12,6 +12,9 @@ import (
 	"os"
 )
 
+var tv bool = true
+var fv bool = false
```
kidk, line 16: I don't think you use `fv` anywhere.

mbazhlekova: Removed

kidk (approved):

PR #933, opened 2022-09-13
fix: omit filter ID when empty

YentlFrickx:
This is a fix for https://github.com/newrelic/terraform-provider-newrelic/issues/1986 . When trying to update a workflow via terraform an error is thrown due to the fact that no filter ID (empty) is passed the first time. By allowing this to be omitted when empty (which is still a valid request) the terraform update of a workflow works and results in a valid workflow.


kidk (approved):

kidk:
Thanks a lot @YentlFrickx Merging it in.

PR #934, opened 2022-09-15
feat: update servicelevel model

lucasmessinar:
Updating service level model with latest changes to allow the usage of sum function.


NSSPKrishna (dismissed):

lucasmessinar:
This is a breaking change, since now the function requires passing the GUID of the SLI as well.

mbazhlekova (approved):

PR #937, opened 2022-10-10
(fix)  address workflow update marshalling issues

IliaShurygin:
**THIS IS A BREAKING CHANGE** (see note at the bottom)

This PR addresses a number of issues with the API used to update workflows
1. It was impossible to update any boolean field from `true` to `false`
2. Empty `enrichments` and `destinationConfigurations` fields were not marshalled correctly
3. String fields with empty values were not marshalled correctly
4. Empty filter was not marshalled correctly

In addition to the fixes themselves, this PR adds a lot of additional tests and assertions for both the "update" as well as other workflows APIs

Unfortunately, it was impossible to implement these changes in a backward-compatible way. The current API is just does not match our GraphQL API


NSSPKrishna (dismissed):

mbazhlekova:
@IliaShurygin Looks good! Just a few things:

1. Can you edit your commit messages to have a `:` after the change type? Like this `fix(workflows) make boolean fields optional` -> `fix(workflows): make boolean fields optional` The linter is very strict :) 
2. To mark something as a breaking change, you'll need to modify one commit message to one of these formats:
```
fix: fixed thing xyz BREAKING CHANGE: this will break users because of blah
 
fix!: fixed something 

feat!: added blah
```
This will let our auto-releaser know to bump a major version.

mbazhlekova (approved):

NSSPKrishna (approved):

mbazhlekova:
Hi @IliaShurygin have some more linter fun 🙂

This commit - chore(workflows) rework integration tests - is missing the `:` after `chore(workflows)` which is upsetting the linter

mbazhlekova (approved):

PR #940, opened 2022-10-18
NR-52189: added missing userdef values

NSSPKrishna:
## Why?
The mobile_alerts in the alert conditions also supports percent and rate for user defined value function.


kidk (approved):

mbazhlekova (approved):

mbazhlekova:
@NSSPKrishna Can you make this a `feat:` or `fix:` commit message so it will trigger a release?

mbazhlekova (approved):

Issue #932, opened 2022-09-08
Adding "omitempty" will prevent boolean field from being sent to API if false


usfdmiller:
### Description
If a boolean field in a struct defines an "omitempty" option, that field will always be omitted when encoded to a JSON object if the value is false.  This is a problem with the workflow resource as it is essentially impossible to disable the workflow due to the workflow_enabled field being omitted.  I'm certain this issue is not limited to workflows as there appear to be multiple boolean fields that define this option.  

### Go Version
go version go1.19.1 darwin/amd64

### Current behavior
Any boolean field in a struct with the "omitempty" option will be omitted from the JSON encoding if the value is false.   

### Expected behavior
As false is a perfectly valid value for a boolean, I would expect a boolean field with a false value not to be omitted from the JSON encoding. 

### Steps To Reproduce
1. Create a Terraform configuration that defines a workflow.
2. Ensure the "workflow_enabled" attribute has a value of "false"
3. Run "terraform apply" to create the resources.
4. Notice how the workflow is not actually disabled.

### Debug Output (if applicable)
I don't have any debug output, but debug output from any workflow creation should demonstrate a false value results in the field being omitted.

### Additional Context
I'm not a GO expert, so I'm basing this on the documentation for the json package.

https://pkg.go.dev/encoding/json#Marshal

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string. 

I would attempt to submit a pull request, but it appears the workflow package was generated with a tool, and I was only able to remove the "omitempty" option manually.  

### References or Related Issues
Probably the root cause for [issue #1964](https://github.com/newrelic/terraform-provider-newrelic/issues/1964) in the provider repository.

PR #943, opened 2022-10-21
fix(workflows)!: updated how empty fields are handled on create

usfdmiller:
_**This is a breaking change.**_  I'm unabashedly copying the changes made to AiWorkflowsUpdateWorkflowInput in pull request #937 and applying them to AiWorkflowsCreateWorkflowInput.  I'm hoping this aligns workflow behavior between creates and updates around empty fields and JSON marshaling.  I'd _really_ like to create workflows in a disabled state and this is one of the last little issues preventing me from upgrading to the 3.0+ provider.  I even created issue #932 a while back regarding this problem.

This is pushing my novice GO knowledge, so please let me know if I've made a questionable change, I'm not meeting standards, or if I've left out information.    

Thanks!


mbazhlekova:
@usfdmiller Thank you for your contribution! Sorry for the delay on this. We're doing some feature work that we're hoping to get into v2 before releasing another major version since it's a bit of a hassle in Go. Happy to review once that work is finished.

usfdmiller:
It looks like my core issue here, being unable to create workflows in a disabled state, was resolved in PR #959.   That said, I'm going to close this PR.  Thanks!

PR #950, opened 2022-11-08
(fix) made boolean omitempty values references

amaor-newrelic:
**THIS IS A BREAKING CHANGE** 
This PR addresses a number of issues with the API used to update channels and destinations
1. It was impossible to update any boolean field from true to false
4. Empty filter was not marshalled correctly
Unfortunately, it was impossible to implement these changes in a backward-compatible way. The current API is just does not match our GraphQL API.

See [this pr](https://github.com/newrelic/newrelic-client-go/pull/937) for reference


mbazhlekova:
@amaor-newrelic Thank you for your contribution! Sorry for the delay on this. We're doing some feature work that we're hoping to get into v2 before releasing another major version since it's a bit of a hassle in Go. Happy to review once that work is finished.

PR #962, opened 2022-11-30
feat(dashboard): update dashboard entity query with variables

mbazhlekova:


pkg/dashboards/dashboard.go
```diff
@@ -87,6 +87,26 @@ const getDashboardEntityQuery = `query ($guid: EntityGuid!) {
         tags { key values }
         tagsWithMetadata { key values { mutable value } }
         updatedAt
+		variables {
```
sanderblue, line 90: Indention seems off

sanderblue (dismissed):

Issue #982, opened 2023-02-03
Error creating synthetic script monitor targeting legacy runtime.


usfdmiller:
### Description
Trying to create a newrelic_synthetics_script_monitor resource targeting the legacy (Node 10) runtime results in an error.  

### Go Version
go version go1.19.4 darwin/amd64

### Current behavior
As the description say, trying to create a synthetic script monitor that targets the legacy Node 10 runtime results in an error.

```json
{
 "data": {
  "syntheticsCreateScriptApiMonitor": {
   "errors": [
    {
     "description": "Runtime values are invalid combination.",
     "type": "BAD_REQUEST"
    }
   ],
   "monitor": null
  }
 }
}
```

### Expected behavior
I expect the synthetic script monitor to be deployed without error.  According to the documentation it should still be possible to target the legacy runtime.

> If you want to use a legacy runtime (Node 10 or Chrome 72) you can set the runtime_type, runtime_type_version and script_language to empty string "". 

### Steps To Reproduce

1. Run "terraform apply" with the following configuration.

```hcl
resource "newrelic_synthetics_script_monitor" "monitor" {
  status               = "ENABLED"
  name                 = "script_monitor"
  type                 = "SCRIPT_API"
  locations_public     = ["US_WEST_1"]
  period               = "EVERY_6_HOURS"

  script               = "console.log('it works!')"

  script_language      = ""
  runtime_type         = ""
  runtime_type_version = ""

  tag {
    key    = "some_key"
    values = ["some_value"]
  }
}
```

### Debug Output (if applicable)
Looking the debug output for the request, the script_language field is not being sent in the API payload.

```json
{
 "query": "mutation(\n\t$accountId: Int!,\n\t$monitor: SyntheticsCreateScriptApiMonitorInput!,\n) { syntheticsCreateScriptApiMonitor(\n\taccountId: $accountId,\n\tmonitor: $monitor,\n) {\n\terrors {\n\t\tdescription\n\t\ttype\n\t}\n\tmonitor {\n\t\tcreatedAt\n\t\tguid\n\t\tid\n\t\tlocations {\n\t\t\tprivate\n\t\t\tpublic\n\t\t}\n\t\tmodifiedAt\n\t\tname\n\t\tperiod\n\t\truntime {\n\t\t\truntimeType\n\t\t\truntimeTypeVersion\n\t\t\tscriptLanguage\n\t\t}\n\t\tstatus\n\t}\n} }",
 "variables": {
[...]

PR #983, opened 2023-02-03
fix(synthetics): resolved error targeting legacy runtimes

usfdmiller:
The _omitempty_ directive is preventing the ScriptLanguage field from being passed to the API when set to an empty string.  According to the documentation an empty string must be specified when targeting the legacy (Node 10) runtime.  This is currently preventing synthetic script monitors that target the legacy runtime from being created.

There is a bit more detail provided in issue #982.  

This pull request includes a fix to remove the _omitempty_ directive, two copied and pasted tests to verify, and a missing integration tag on an unrelated integration test.

### DIFF
--- a/pkg/synthetics/types.go
+++ b/pkg/synthetics/types.go
@@ -808,7 +808,7 @@ type SyntheticsRuntimeInput struct {
 	// The specific version of the runtime type selected
 	RuntimeTypeVersion SemVer `json:"runtimeTypeVersion"`
 	// The programing language that should execute the script
-	ScriptLanguage string `json:"scriptLanguage,omitempty"`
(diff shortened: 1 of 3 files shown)


pkg/synthetics/types.go
```diff
@@ -808,7 +808,7 @@ type SyntheticsRuntimeInput struct {
 	// The specific version of the runtime type selected
 	RuntimeTypeVersion SemVer `json:"runtimeTypeVersion"`
 	// The programing language that should execute the script
-	ScriptLanguage string `json:"scriptLanguage,omitempty"`
+	ScriptLanguage string `json:"scriptLanguage"`
```
sanderblue, line 811: 👍

PR #991, opened 2023-02-22
feat(agentapplication): add ability to manage browser applications

sanderblue:
Related to: [terraform-provider-newrelic](https://github.com/newrelic/terraform-provider-newrelic/issues/73)

Note: We currently can't return `apmSettings.alias` (this was manually removed from the mutation) in the `agentApplicationSettingsUpdate` response due to a [known error resolving that field](https://issues.newrelic.com/browse/NR-95551) on the API side. We can probably move forward without this for the time being.


rthorn-nr (approved):

rthorn-nr (approved):

pkg/testhelpers/helpers.go
```diff
@@ -51,3 +51,17 @@ func getEnvInt(name string) (int, error) {
...
 	return n, nil
 }
+
+// Use this method to generate a random name to be used in integration tests
+// for whatever resource you might be trying to create and/or test.
+// If no random character count is provided, the default behavior is to append 5 random
+// letters to the end of the name.
+//
+// Example random name: nr-test-xmnvb
+func GenerateRandomName(randCharCount int) string {
```
Julien4218, line 61: @sanderblue this seems to only be used in a single test file, perhaps this should move in that test file?

sanderblue: I'll be updating other tests to use it shortly 😉

PR #999, opened 2023-03-06
fix(EpochTime) fix EpochTimeUnix unit test

diabeticMike:
One unit test is failing for my previous PR:https://github.com/newrelic/newrelic-client-go/pull/997
So i made fix for that
Failed CI: https://github.com/newrelic/newrelic-client-go/actions/runs/4345000784/jobs/7589103597
@sanderblue

PR #1000, opened 2023-03-06
chore: fix epoch time unit test for empty time

sanderblue:
This PR also include dependency upgrades for some security vulnerability warnings we received via `dependabot`.


diabeticMike:
Hi @sanderblue 
i made same PR here ))
https://github.com/newrelic/newrelic-client-go/pull/999

sanderblue:
Note: failing data partition tests have to do with our tests using our old v1 test account vs our new v2 test account.

PR #1007, opened 2023-03-08
chore(new_test_account): updated guids accountId applicationID with new account settings

RavitejaSurampudi:
Updated  below properties for the new account
1. new account id
2. updated all guids  for the new dummy app created 
3. updated application id for the app and mobile


sanderblue:
Fixed in #1010

PR #1012, opened 2023-03-16
feat: Update servicelevel schema

asanuy:
Changes include support to configure SLIs using the GET_FIELD and GET_CDF_COUNT functions. Additionally, it has been introduced a new threshold property that is necessary for CDF.

### DIFF
--- a/pkg/servicelevel/types.go
+++ b/pkg/servicelevel/types.go
@@ -11,14 +11,22 @@ import (
 type ServiceLevelEventsQuerySelectFunction string
 
 var ServiceLevelEventsQuerySelectFunctionTypes = struct {
-	// COUNT.
+	// Use on events and unaggregated data.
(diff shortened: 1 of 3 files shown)


pkg/servicelevel/types.go
```diff
@@ -303,6 +319,16 @@ type ServiceLevelResultQuery struct {
 	NRQL NRQL `json:"nrql"`
 }
 
+// User - The `User` object provides general data about the user.
+type User struct {
```
asanuy, line 323: This struct is unrelated to the changes on the API. Is this new object expected and harmless?

sanderblue: In this case it's probably harmless. Not quite sure why it was generated and added to the `Actor` struct, but sometimes Tutone might generate an extraneous type to satisfy the GraphQL schema.

PR #1014, opened 2023-03-22
fix(notifications): add name to filter

lzaga-newrelic:
Add to AiNotificationsDestinationFilter the Name attribute.

### DIFF
--- a/pkg/notifications/notifications_integration_test.go
+++ b/pkg/notifications/notifications_integration_test.go
@@ -53,7 +53,7 @@ func TestNotificationMutationDestination(t *testing.T) {
 	require.NotEmpty(t, createResult.Destination.Auth)
 	require.Equal(t, ai.AiNotificationsAuthType("TOKEN"), createResult.Destination.Auth.AuthType)
 
-	// Test: Get Destination
+	// Test: Get Destination by id
 	filters := ai.AiNotificationsDestinationFilter{
 		ID: createResult.Destination.ID,
 	}
@@ -62,6 +62,15 @@ func TestNotificationMutationDestination(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, getDestinationResult)
 	assert.Equal(t, 1, getDestinationResult.TotalCount)
+
+	// Test: Get Destination by name
+	filtersByName := ai.AiNotificationsDestinationFilter{
+		Name: createResult.Destination.Name,
+	}
+	getDestinationByNameResult, err := n.GetDestinations(accountID, "", filtersByName, sorter)
+	require.NoError(t, err)
+	require.NotNil(t, getDestinationByNameResult)
+	assert.Equal(t, 1, getDestinationByNameResult.TotalCount)
 
 	// Test: Update Destination
 	updateDestination := AiNotificationsDestinationUpdate{}
--- a/pkg/notifications/notifications_integration_test.go
+++ b/pkg/notifications/notifications_integration_test.go
@@ -15,7 +15,7 @@ import (
 	mock "github.com/newrelic/newrelic-client-go/v2/pkg/testhelpers"
 )
 
-func TestNotificationMutationDestinationById(t *testing.T) {
+func TestNotificationMutationDestination(t *testing.T) {
 	t.Parallel()
 
 	n := newIntegrationTestClient(t)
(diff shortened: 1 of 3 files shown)


pranav-new-relic:
Looks good to me 👍

pranav-new-relic (approved):

pkg/notifications/notifications_integration_test.go
```diff
@@ -63,6 +63,15 @@ func TestNotificationMutationDestination(t *testing.T) {
 	require.NotNil(t, getDestinationResult)
 	assert.Equal(t, 1, getDestinationResult.TotalCount)
 
+	// Test: Get Destination by name
+	filtersByName := ai.AiNotificationsDestinationFilter{
```
IliaShurygin, line 67: Could you please extract this into a separate test? 
With the current setup, if the destination filtering logic would break, it would break the test called `TestNotificationMutationDestination`, which is quite misleading

lzaga-newrelic: done!

pranav-new-relic (dismissed):

pkg/notifications/notifications_integration_test.go
```diff
@@ -93,6 +93,84 @@ func TestNotificationMutationDestination(t *testing.T) {
 	require.NotNil(t, deleteResult)
 }
 
+func TestNotificationMutationDestinationByName(t *testing.T) {
```
IliaShurygin, line 96: We are just testing data source, not the update mutation.
I would suggest to both rename the test (to, e.g., `TestNotificationGetDestinationByName`), as well as remove  the irrelevant parts from it (see comments below)

pkg/notifications/notifications_integration_test.go
```diff
@@ -93,6 +93,84 @@ func TestNotificationMutationDestination(t *testing.T) {
...
+		},
+	}
+	destination.Name = fmt.Sprintf("test-notifications-destination-%s", testIntegrationDestinationNameRandStr)
+
+	// Test: Create
+	createResult, err := n.AiNotificationsCreateDestination(accountID, destination)
+	require.NoError(t, err)
+	require.NotNil(t, createResult)
+	require.NotEmpty(t, createResult.Destination.Auth)
+	require.Equal(t, ai.AiNotificationsAuthType("TOKEN"), createResult.Destination.Auth.AuthType)
```
IliaShurygin, line 132: nitpick: these parts are already verified in the previous test, there's no point to check it again. It just makes reading the test harder and creates additional maintenance burden

pkg/notifications/notifications_integration_test.go
```diff
@@ -93,6 +93,84 @@ func TestNotificationMutationDestination(t *testing.T) {
...
+		Token: AiNotificationsTokenAuthInput{
+			Token:  "TokenUpdate",
+			Prefix: "BearerUpdate",
+		},
+	}
+	updateDestination.Name = fmt.Sprintf("test-notifications-update-destination-%s", testIntegrationDestinationNameRandStr)
+
+	updateDestinationResult, err := n.AiNotificationsUpdateDestination(accountID, updateDestination, createResult.Destination.ID)
+	require.NoError(t, err)
+	require.NotNil(t, updateDestinationResult)
```
IliaShurygin, line 166: Destination updates after fetching it by name are completely irrelevant for the test

pkg/notifications/notifications_integration_test.go
```diff
@@ -93,6 +93,84 @@ func TestNotificationMutationDestination(t *testing.T) {
...
+
+	// Test: Delete
+	deleteResult, err := n.AiNotificationsDeleteDestination(accountID, createResult.Destination.ID)
+	require.NoError(t, err)
+	require.NotNil(t, deleteResult)
```
IliaShurygin, line 171: What if the test fails? Do we not remove the destination?

lzaga-newrelic: Not sure there is a better way of handling this - also I checked other places in this repo and it doesn't seems anyone else is cleaning this. @pranav-new-relic do you know if there is a cleaning after the integration tests for all the resources that weren't deleted?

pranav-new-relic: I don't think so, but I think maybe what @IliaShurygin means is the usage of cleanup mechanisms like in the Workflows Integration Test (I found the following as an example that you had both written)

https://github.com/newrelic/newrelic-client-go/blob/main/pkg/workflows/workflows_integration_test.go#L519

Would you like to take a look and see if something similar can be added in the current test when there is an error thrown (and is unexpected, which would eventually lead to a failure)

lzaga-newrelic: Actually, the cleanup is another call to the API (which originally failed in the test), so I'm not sure how this should help other than it has one more attempt to delete it...

pranav-new-relic (approved):

PR #1034, opened 2023-07-07
fix(data_partition): fix issue with enable attribute

RavitejaSurampudi:
fix issue with enable attribute


NSSPKrishna (approved):
Looks good, can you please include a comment for the developer to not generate using tutone or re-edit so it stays inline with the TF?

pranav-new-relic (approved):

PR #1038, opened 2023-07-19
chore(apiaccess): Updated CreatedAt attr to nrtime.EpochSeconds

mbruzina:
Changed EpochSeconds to nrtime.EpochSeconds.  This appears to be the correct golang time to handle `createdAt` fields returned in NerdGraph responses

Many of the files touched were a result of running `go fmt` to satisfy linter checks

### DIFF
--- a/pkg/apiaccess/keys.go
+++ b/pkg/apiaccess/keys.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
 	"github.com/newrelic/newrelic-client-go/v2/internal/http"
 )
 
@@ -220,6 +221,7 @@ const (
 		id
 		key
 		name
+		createdAt
 		notes
 		type
 		... on ApiAccessIngestKey {
--- a/pkg/apiaccess/types.go
+++ b/pkg/apiaccess/types.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/newrelic/newrelic-client-go/v2/pkg/accounts"
+	"github.com/newrelic/newrelic-client-go/v2/pkg/nrtime"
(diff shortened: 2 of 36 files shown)


pkg/apiaccess/types.go
```diff
@@ -577,8 +578,15 @@ type APIAccessUserKeyError struct {
 
 func (x *APIAccessUserKeyError) ImplementsAPIAccessKeyError() {}
 
-// EpochSeconds - The `EpochSeconds` scalar represents the number of seconds since the Unix epoch
-type EpochSeconds string
+type APIAccessKeyErrorResponse struct {
```
mbruzina, line 581: This seems to have been added manually originally, needed to do the same after `tutone generate` removed

pkg/apiaccess/types.go
```diff
@@ -334,12 +335,12 @@ func (x *APIAccessIngestKeyError) ImplementsAPIAccessKeyError() {}
 // APIAccessKey - A key for accessing New Relic APIs.
 type APIAccessKey struct {
 	// The UNIX epoch when the key was created, in seconds.
-	CreatedAt EpochSeconds `json:"createdAt,omitempty"`
+	CreatedAt nrtime.EpochSeconds `json:"createdAt,omitempty"`
```
mbruzina, line 338: All of these `CreatedAt` attributes used to reference `EpochSeconds`, a string

pkg/apiaccess/keys.go
```diff
@@ -220,6 +221,7 @@ const (
 		id
 		key
 		name
+		createdAt
```
mbruzina, line 224: update response fields to return the `createdAt` time

PR #1055, opened 2023-08-23
chore(synthetics): addition of schema and functions for synthetics automated tests

pranav-new-relic:
This PR includes 
- changes fetched by Tutone in the **synthetics** package to enable the query `automatedTestResults` and the mutation `syntheticsStartAutomatedTest` and associated types and details.
- tests covering multiple scenarios of the mutation and the query

The changes made have nearly been integrated with the CLI, and have been tested with multiple combinations to ensure the properties of attributes generated and listed in this PR lead to no unexpected behaviour in requests unmarshaled from YAML/responses parsed.

**Note:**
- Except for the removal of the `omitempty` of two fields `IsBlocking` and `Overrides` generated by Tutone, no other manual changes have been made to the code generated. All other datatype/struct tag overrides are as specified and generated from `tutone.yml`.
- The added tests are expected to fail in the CI owing to the fact that the query and mutation are not yet enabled to all customers, and the account that was used to fetch the schema via Tutone is not used in the CI.

### DIFF
--- a/pkg/synthetics/synthetics_api_integration_test.go
+++ b/pkg/synthetics/synthetics_api_integration_test.go
@@ -4,12 +4,14 @@
 package synthetics
 
 import (
+	"errors"
+	"fmt"
+	"log"
+	"os"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
-
-	"fmt"
-	"os"
 
 	mock "github.com/newrelic/newrelic-client-go/v2/pkg/testhelpers"
 )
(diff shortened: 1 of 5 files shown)


pkg/synthetics/types.go
```diff
@@ -502,12 +590,142 @@ func (x *Nr1CatalogTargetedInstallPlanDirective) ImplementsNr1CatalogInstallPlan
...
+	// Specific overrides for the given monitor
+	Overrides SyntheticsAutomatedTestOverrides `json:"overrides,omitempty"`
+}
+
+// SyntheticsAutomatedTestMonitorConfigInput - Monitor specific test configuration
+type SyntheticsAutomatedTestMonitorConfigInput struct {
+	// Specifies whether a failure of this monitor should fail the entire automated test
+	IsBlocking bool `json:"isBlocking" yaml:"isBlocking"`
+	// Specific overrides for the given monitor
+	Overrides *SyntheticsAutomatedTestOverridesInput `json:"overrides" yaml:"overrides"`
```
pranav-new-relic, line 678: `omitEmpty` manually discarded from these attributes, because
- if `isBlocking` is given `false` in the YAML file, `isBlocking = false` is not even sent in the request to the API, which does not affect the response of the API, but would seem unexpected to anyone debugging this.
- if no `overrides` are specified in the YAML file, all contents of the `overrides` object are created as empty strings `""` which is sometimes leading to errors; hence, a pointer `*` has been added. However, with `omitEmpty`, if no `overrides` are specified in the input, it still disappears from the API request, so it has been removed.

pranav-new-relic:
Rebased this branch onto main - a recent PR merged to `main` yesterday includes changes in the `synthetics` package **not** associated with the Automated Tests mutation and query; hence, upon rebasing, this PR now includes changes specific only to the Automated Tests mutation and query (and does not include other changes from the `synthetics` package).

pkg/synthetics/synthetics_api_integration_test.go
```diff
@@ -1022,3 +1024,345 @@ func newIntegrationTestClient(t *testing.T) Synthetics {
...
 	return New(tc)
 }
+
+// TestSyntheticsStartAutomatedTest_Basic performs a test by creating three monitors and using the
+// syntheticsStartAutomatedTest mutation to create a batch with these three monitors. The expected
+// behaviour of this test is to return a valid batchId and throw no error.
+func TestSyntheticsStartAutomatedTest_Basic(t *testing.T) {
+	t.Parallel()
+	t.Skipf(
+		`Temporarily skipping tests associated with the Synthetics Automated Tests feature, ` +
```
Nandu-pns, line 1034: Good to proceed for now, hoping we shall take a note of re-testing once API is stable.

pranav-new-relic: Right - these tests are currently being skipped, given that the API is not producing expected results (fully) on production yet, and is a bit more stable on staging, but is still erroneous at times, leading to test failures of some of these, such as the unit tests written. However, all of these tests have been locally verified, and have been written in alignment with the expected behaviour of the API, to ensure these tests can be unskipped once the API is stable.

PR #1065, opened 2023-11-13
chore(events): add events api

Julien4218:


newrelic/events_service.go
```diff
@@ -0,0 +1,14 @@
+package newrelic
+
+import (
+	"context"
+)
+
+type EventsAPI interface {
```
sanderblue, line 7: For our future selves, can you add a doc block above this that explains what this is and why it's here please? 🙂

Julien4218: done

sanderblue (approved):

sanderblue (approved):

PR #1067, opened 2023-11-15
chore: make names of synthetics entities created in tests consistent

pranav-new-relic:
In multiple tests used in the `synthetics` package which end up creating/updating/deleting Synthetic Monitors, Private Locations and Secure Credentials, the names of resources created are currently not consistent - 
- Very few of them have a string that helps identify that the source of the entity is `newrelic-client-go`.
- A large number of them have no such identifier and simply have a `mock.RandSeq(5)` as the name of the entity.

This makes it difficult to track down redundant Synthetic entities in the account which may have not been deleted successfully owing to test failures or other similar reasons, and which would need to be cleaned up (given, in recent times, a large number of redundant monitors created by the client's test which remained and did not get deleted, led to the limit of Synthetic Checks getting breached).

Hence, this PR addresses a tiny change to all tests in the `synthetics` package - make the names of entities created uniform, with a identifier `client-go-test` that helps identify that the entities were created from `newrelic-client-go`'s tests.

### DIFF
--- a/pkg/entities/entity_integration_test.go
+++ b/pkg/entities/entity_integration_test.go
@@ -257,7 +257,7 @@ func TestIntegrationGetEntity_MobileEntity(t *testing.T) {
 
 func TestIntegrationGetEntity_SyntheticsEntity(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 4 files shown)


pranav-new-relic:
All tests successful 👍 however, it is worth mentioning (for anyone seeing this PR in future) that one specific test, `TestIntegrationListTags` is very flakey - locally running this tests seems to cause no failure, but this fails in the CI quite frequently, and then, randomly passes in the next one or two test runs.

Merging this PR shortly as this is a minor change (expected to cause no change to feature behaviour) and since all tests have passed successfully.

PR #1070, opened 2023-12-12
feat(synthetics): addition of mutations for the monitor downtime feature

pranav-new-relic:
#### Summary

This PR includes an addition of create, edit and delete mutations for the Synthetics Monitor Downtime feature. All types and mutations have been fetched via Tutone, however, the `Edit` mutations have been shifted out of the scope of Tutone, owing to a blocker caused by the structure of the mutation, and thereby, split into four - one each for each type of downtime, similar to the create mutations.

#### Tests

Around 10 tests have been added to cover all success and most failure possibilities of create, edit and delete requests and responses with Monitor Downtimes, and these are seen to pass in the CI run.


Nandu-pns:
Changes looks good to me.

Nandu-pns (approved):

PR #1077, opened 2024-01-18
fix: When using a testserver, ensure the Insights API is mocked

joshbranham:
We were using the `testhelpers` package to setup a `NewTestConfig` and noticed when trying to test the Insights API calls, it was not being mocked. I believe this is not intentional, so this is a quick fix, but please let me know if this is intentional.

Fixes https://github.com/newrelic/newrelic-client-go/issues/1107


pranav-new-relic (approved):

pranav-new-relic:
looks good to me 👍 thank you for the contribution

PR #1086, opened 2024-01-25
feat(authentication_domains): addition of client code to fetch auth domains

pranav-new-relic:
#### Jira
https://new-relic.atlassian.net/browse/NR-76861

#### Summary
This PR aims at adding the following - 
- Addition of a query to fetch `authenticationDomains`, which would be a prerequisite for a data source we're planning to add in the Terraform Provider to fetch Authentication Domains
   - We're adding this in order to be able to use the data source to fetch the ID of the authentication domain to use with (upcoming) group management and user management resources 
- Scaffolding a new package `usermanagement` that would contain queries and mutations to manage authentication domains (only read), users and groups for the above
- Addition of Tutone generated types for the above (generated for the read query added to the Tutone YML file)
- Addition of unit and integration tests to validate the query added


Nandu-pns (approved):

pranav-new-relic:
The integration test failures are due to an entity that was active so far (but is no longer reporting) which we're trying to revive - the `usermanagement` tests added so far, however, work fine - so we should be good to merge this.

image

PR #1091, opened 2024-02-07
chore(tests): fix tests by replacing obsolete entities and GUIDs

pranav-new-relic:
Fixes to a few integration tests which are currently failing to run because GUIDs of entites they use no longer exist.
More details on changes made may be found in the comments added against each test-case change.

### DIFF
--- a/pkg/entities/entity_integration_test.go
+++ b/pkg/entities/entity_integration_test.go
@@ -140,10 +140,8 @@ func TestIntegrationGetEntities(t *testing.T) {
 func TestIntegrationGetEntity(t *testing.T) {
 	t.Parallel()
 
-	// GUID of Dummy App
-	// entityGUID := common.EntityGUID(testhelpers.IntegrationTestApplicationEntityGUID)
-
-	// Temporarily using the GUID of a different app
+	// GUID of 'Dummy App Pro Max', a replacement to 'Dummy App' (testhelpers.IntegrationTestApplicationEntityGUID)
+	// as Dummy App is no longer reporting
 	entityGUID := common.EntityGUID("MzgwNjUyNnxBUE18QVBQTElDQVRJT058NTUzNDQ4MjAy")
 
 	client := newIntegrationTestClient(t)
@@ -164,7 +162,6 @@ func TestIntegrationGetEntity(t *testing.T) {
 	assert.Equal(t, "APM", actual.Domain)
 	assert.Equal(t, EntityType("APM_APPLICATION_ENTITY"), actual.EntityType)
 	assert.Equal(t, entityGUID, actual.GUID)
-	//assert.Equal(t, "Dummy App", actual.Name)
 	assert.Equal(t, "Dummy App Pro Max", actual.Name)
 	assert.Equal(t, "https://one.newrelic.com/redirect/entity/"+string(entityGUID), actual.Permalink)
 	assert.Equal(t, true, actual.Reporting)
@@ -176,10 +173,8 @@ func TestIntegrationGetEntity_ApmEntity(t *testing.T) {
 
 	client := newIntegrationTestClient(t)
 
-	// GUID of Dummy App
-	// result, err := client.GetEntity(testhelpers.IntegrationTestApplicationEntityGUID)
-
-	// Temporarily using the GUID of a different app
+	// GUID of 'Dummy App Pro Max', a replacement to 'Dummy App' (testhelpers.IntegrationTestApplicationEntityGUID)
+	// as Dummy App is no longer reporting
 	result, err := client.GetEntity("MzgwNjUyNnxBUE18QVBQTElDQVRJT058NTUzNDQ4MjAy")
 
 	if e, ok := err.(*http.GraphQLErrorResponse); ok {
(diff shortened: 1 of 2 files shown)


pkg/entities/entity_integration_test.go
```diff
@@ -141,7 +141,11 @@ func TestIntegrationGetEntity(t *testing.T) {
 	t.Parallel()
 
 	// GUID of Dummy App
-	entityGUID := common.EntityGUID(testhelpers.IntegrationTestApplicationEntityGUID)
+	// entityGUID := common.EntityGUID(testhelpers.IntegrationTestApplicationEntityGUID)
+
+	// Temporarily using the GUID of a different app
+	entityGUID := common.EntityGUID("MzgwNjUyNnxBUE18QVBQTElDQVRJT058NTUzNDQ4MjAy")
+
```
pranav-new-relic, line 148: `func TestIntegrationGetEntity(t *testing.T)`
- _temporarily_ changed the GUID of the entity being used by the test to fetch details of, which is a Python-based APM entity, `Dummy App Pro Max`
- accordingly, changed assert statements in the following lines of the same function to match the ID of the entity, language, etc.

pkg/entities/entity_integration_test.go
```diff
@@ -172,7 +177,10 @@ func TestIntegrationGetEntity_ApmEntity(t *testing.T) {
 	client := newIntegrationTestClient(t)
 
 	// GUID of Dummy App
-	result, err := client.GetEntity(testhelpers.IntegrationTestApplicationEntityGUID)
+	// result, err := client.GetEntity(testhelpers.IntegrationTestApplicationEntityGUID)
+
+	// Temporarily using the GUID of a different app
+	result, err := client.GetEntity("MzgwNjUyNnxBUE18QVBQTElDQVRJT058NTUzNDQ4MjAy")
```
pranav-new-relic, line 183: `func TestIntegrationGetEntity_ApmEntity(t *testing.T)`
- _temporarily_ changed the GUID of the entity being used by the test to fetch details of, which is a Python-based APM entity, `Dummy App Pro Max`
- accordingly, changed assert statements in the following lines of the same function to match the ID of the entity, language, etc.

pkg/entities/entity_integration_test.go
```diff
@@ -238,20 +248,24 @@ func TestIntegrationGetEntity_MobileEntity(t *testing.T) {
...
+		t.Skipf("Skipping this test as MobileApplicationEntities are fragile, need to be recreated=")
+	}
+
+	require.NotNil(t, (*result))
 	actual := (*result).(*MobileApplicationEntity)
 
 	// These are a bit fragile, if the above GUID ever changes...
 	// from MobileApplicationEntity / MobileApplicationEntityOutline
-	assert.Equal(t, 601375901, actual.ApplicationID)
+	assert.Equal(t, 601445161, actual.ApplicationID)
```
pranav-new-relic, line 268: `func TestIntegrationGetEntity_MobileEntity(t *testing.T)`
- since the older GUID of the Mobile Entity (which has been in use so far) is obsolete and can no longer be found in the New Relic account, _temporarily_ changed the GUID to that of a new Mobile Entity
- however, since instrumented mobile entities are fragile too (as we don't follow a regular APM-type procedure), added a `Skip` method to not report a failure, if the Mobile Entity with this GUID ceases to exist some day

pkg/entities/entity_integration_test.go
```diff
@@ -238,20 +248,24 @@ func TestIntegrationGetEntity_MobileEntity(t *testing.T) {
...
+		t.Skipf("Skipping this test as MobileApplicationEntities are fragile, need to be recreated")
+	}
+
+	require.NotNil(t, (*result))
 	actual := (*result).(*MobileApplicationEntity)
 
 	// These are a bit fragile, if the above GUID ever changes...
 	// from MobileApplicationEntity / MobileApplicationEntityOutline
-	assert.Equal(t, 601375901, actual.ApplicationID)
+	assert.Equal(t, 601445161, actual.ApplicationID)
```
pranav-new-relic, line 268: `func TestIntegrationGetEntity_MobileEntity(t *testing.T)`
- since the older GUID of the Mobile Entity (which has been in use so far) is obsolete and can no longer be found in the New Relic account, _temporarily_ changed the GUID to that of a new Mobile Entity
- however, since instrumented mobile entities are fragile too (as we don't follow a regular APM-type procedure), added a `Skip` method to not report a failure, if the Mobile Entity with this GUID ceases to exist some day

pkg/entities/tags_integration_test.go
```diff
@@ -31,28 +31,7 @@ func TestIntegrationListTags(t *testing.T) {
...
-	require.NoError(t, err)
-	require.Greater(t, len(actual), 0)
-
-	actual, err = client.GetTagsForEntityMutable(testGUID)
-
-	require.NoError(t, err)
-	require.Greater(t, len(actual), 0)
-}
-
-func TestIntegrationTaggingAddTagsToEntity(t *testing.T) {
```
pranav-new-relic, line 55: `func TestIntegrationGetTagsForEntity(t *testing.T)`
- This test is failing often (though not always) because of a probable consequence of parallel execution. Tests in the same file listed below this test add, update and delete tags; so it could be that whenever this test fails, it is happening because this test runs after the added tags are deleted by other tests, and hence, the failure to "fetch" any existing tags.
- As a probable workaround, I felt merging both of these tests might make sense - tags would be added, and immediately, fetched, to see if they've actually been added.

pranav-new-relic:
Tests passing on other PRs too 👍 (pasting this for future reference)

image

PR #1103, opened 2024-02-28
fix(group_management): alter group management queries and add tests

pranav-new-relic:
TBD


pkg/synthetics/synthetics_api_integration_test.go
```diff
@@ -796,7 +796,8 @@ func TestSyntheticsCertCheckMonitor_Basic(t *testing.T) {
 				Values: []string{"avocado"},
 			},
 		},
-		Domain:                            "https://www.google.com",
+		// do not add a "https://" to the domain; recent error handling in Synthetics throws an "invalid domain" error with "https://"
+		Domain:                            "www.google.com",
```
pranav-new-relic, line 800: With `https://www.google.com`

image

This error does not appear without the `https://`

PR #1122, opened 2024-04-04
feat(destinations): add custom header auth and secureUrl to destinations

nzur-newrelic:
### For Maintainers
⚠️ Kindly hold on merging this PR until 18 April, 2024.

This PR adds 2 new changes to destinations schema:
- new type of auth to destinations called custom headers
- new field secureUrl. This struct holds a SecureValue, so only the prefix part which is a non-SecureValue will be shown upon querying the API.

Required by https://github.com/newrelic/terraform-provider-newrelic/pull/2635/

### DIFF
--- a/pkg/notifications/notifications_api.go
+++ b/pkg/notifications/notifications_api.go
@@ -184,6 +184,12 @@ const AiNotificationsCreateDestinationMutation = `mutation(
 			  authType
 			  prefix
 			}
+			... on AiNotificationsCustomHeadersAuth {
+			  authType
+        	  customHeaders {
+          	    key
+			  }
+			}
 		}
 		createdAt
 		id
@@ -196,6 +202,9 @@ const AiNotificationsCreateDestinationMutation = `mutation(
 			key
 			label
 			value
+		}
+		secureUrl {
+			prefix
 		}
 		status
 		type
@@ -431,7 +440,7 @@ const AiNotificationsUpdateChannelMutation = `mutation(
 		updatedAt
 		updatedBy
 	}
-    error {
+	error {
       ... on AiNotificationsConstraintsError {
         constraints {
           dependencies
@@ -553,6 +562,12 @@ const AiNotificationsUpdateDestinationMutation = `mutation(
 			  authType
 			  prefix
 			}
+			... on AiNotificationsCustomHeadersAuth {
+			  authType
+        	  customHeaders {
+          	    key
+			  }
+			}
 		}
 		createdAt
 		id
@@ -565,6 +580,9 @@ const AiNotificationsUpdateDestinationMutation = `mutation(
 			key
 			label
 			value
+		}
+		secureUrl {
+			prefix
 		}
 		status
 		type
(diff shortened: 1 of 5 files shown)


pkg/notifications/notifications_api.go
```diff
@@ -184,6 +184,12 @@ const AiNotificationsCreateDestinationMutation = `mutation(
 			  authType
 			  prefix
 			}
+			... on AiNotificationsCustomHeadersAuth {
+			  authType
+        	  customHeaders {
+          	    key
+			  }
+			}
```
pranav-new-relic, line 192: image

@nzur-newrelic do we also need to add the attribute `value` here?

pranav-new-relic: Apologies - my bad; I mistook this to be the attributes in the response

nzur-newrelic: No, the part from the screenshot is the parameter, while in the code you're looking at is the response.

nzur-newrelic: @pranav-new-relic

pkg/notifications/types.go
```diff
@@ -714,9 +793,6 @@ type AiNotificationsDestination struct {
 	UpdatedBy int `json:"updatedBy"`
 }
 
-// EntityGUID - An encoded Entity GUID
-type EntityGUID string
-
```
pranav-new-relic, line 719: We might not want to delete this type, as I find that this is being referenced in tests in the Terraform Provider as `notifications.EntityGUID` - please take a look and revert this if necessary.

nzur-newrelic: Yeah that's a mistake, I'll revert it

pranav-new-relic:
@nzur-newrelic these changes look good to me - I've just left a comment and a question. However, let us not merge this PR _until_ the Terraform changes for this are drafted and everything looks good on the Terraform front as well - I mention this, since we've made a lot of pointer and omitempty changes, which are usually a bit fragile. Please let me know once you're done with these changes as well, after which we shall proceed with the remainder of this PR.

pkg/notifications/types.go
```diff
@@ -742,7 +820,9 @@ type AiNotificationsDestinationInput struct {
 	// name
 	Name string `json:"name"`
 	// properties
-	Properties []AiNotificationsPropertyInput `json:"properties,omitempty"`
+	Properties []AiNotificationsPropertyInput `json:"properties"`
```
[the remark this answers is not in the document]
nzur-newrelic: So this is needed, as the type of this on NG is `[AiNotificationsPropertyInput!]!`, it simply has never been tested with an empty value because there is always something there.

pkg/notifications/types.go
nzur-newrelic: Upon taking another look it seems like tutone generate simply moved it to line 1279, kinda weird but it's probably best keeping the what tutone did.

IliaShurygin (approved):

PR #1108, opened 2024-03-11
Authorizationmanagement organization

kporras07:

PR #1100, opened 2024-02-23
feat(usermanagement) Add new getGroups query.

kporras07:

PR #1099, opened 2024-02-22
feat(authorizationmanagement): Add new authorizationmanagement pkg.

kporras07:
Currently only implementing grant as that what we envision to need in the near future.

PR #1098, opened 2024-02-22
feat(job): Add job organization create async result.

kporras07:
Related to #1097 in order to be able to fetch the org create results.

PR #1126, opened 2024-04-10
feat: addition of client code corresponding to packages customeradministration, authorizationmanagement, organization

pranav-new-relic:
#### Background

A new package, `customeradministration` was recently added to NerdGraph (few months ago), after which significant customer interests were received to have these under the purview of the New Relic Go Client, in order to be able to perform these queries in Go. This was conveyed to the team maintaining the Go Client (New Relic Internal: see [this Jira ticket](https://new-relic.atlassian.net/browse/NR-249805) for more details), as a result of which work was logged to have the package added to the client.

In addition to the above, further interests were received in terms of PR contributions, in order to have a few mutations from the `authorizationmanagement` and `organization` packages also included in addition to `customeradministration`, in order to enable a few additional functions for customers via the Go Client (to manage organizations, grants, and more). 

- PRs 
- #1108 
- #1109 
- (formerly #1097 #1098 #1099 #1100)

This PR is a consolidation of all of the above - 
- addition of support to use endpoints within the `customeradministration` package on NerdGraph to perform queries
- addition of support to use _few_ endpoints within the packages `authorizationmanagement` and `organization` to manage grants and organizations
- addition of test cases to cover most of the above

#### Tests
Tests have been added to all relevant code changes, however, a few of these endpoints added have been excluded (owing to reasons [listed here](https://new-relic.atlassian.net/browse/NR-249805?focusedCommentId=744866) - New Relic Internal), to be added as a follow up to this PR after necessary evalutation.

#### Things to Note
This PR also proposes the addition of two new secrets, `INTEGRATION_TESTING_NEW_RELIC_ORGANIZATION_ID` and `INTEGRATION_TESTING_NEW_RELIC_AUTHENTICATION_DOMAIN_ID` as it might not be a good idea to have these included in the code. These secrets have been added to this repository, and require to be added to local environment in order to test most test cases written in this PR.


sanderblue (approved):

pranav-new-relic:
#### ⚠️ A Note About a Limitation These Changes Include
(a general note, for anyone visiting this PR)
##### (As on 16 April, 2024)

In the `customeradministration` package, few types which correspond to fields whose values would need to match either a string, or be contained in a list of strings comprise `Eq string `json:"eq,omitempty"`` and `In []string `json:"in"``, as seen in `customeradministration` > `types.go`. Since the field `Eq` has an `omitempty` tag next to it, while the `In` field does not, failures when trying to use the `Get` methods in this package, with an error stating that these types are expected may be observed. 

This is not a limitation in the New Relic Go Client or with the tool Tutone which we use for code generation; instead, it has been observed that this occurrence is because of the schema definition of these fields in NerdGraph, with `In` as `[ID!]` and `Eq` as `[ID]`, which implies that `In` is a mandatory field with a list of IDs. Though this can be deselected in the NerdGraph UI, since an identical schema exists within the Go Client, the aforementioned failures are seen as `In` is expected to be populated.

It may be noted that adding an `omitempty` tag next to these `In` fields in the Go Client could solve the problem, however, the idea is to engage our friends over at the API team in a conversation to see if this scenario with NerdGraph could be made better, so the Go Client can tune into those changes, without having to deviate from the schema seen on NerdGraph. This shall be an ongoing effort, however, in the interim, in order to avoid the aforementioned error with the specified datatypes comprising `Eq` and `In`, a workaround could be to include the value you would like to specify in the `Eq` field in the `In` field instead, embedded in a list of strings. An example of this would be the following - if you would like to specify 
```go
			AuthenticationDomainId: &MultiTenantAuthorizationGrantAuthenticationDomainIdInputFilter{
				Eq: authenticationDomainId,
			},
```
in order to avoid the aforementioned error, the workaround would be to add the same variable `authenticationDomainId` in a list of strings to `In` instead, as this is currently behaving as a required field.
```go
			AuthenticationDomainId: &MultiTenantAuthorizationGrantAuthenticationDomainIdInputFilter{
				In: []string{
					authenticationDomainId,
				},
			},
```
The following datatypes in `customeradministration` > `types.go` have been observed to display this behaviour and require the aforementioned workaround.			

| Type | Path in NerdGraph |
|--------|--------|
| `MultiTenantAuthorizationGrantAuthenticationDomainIdInputFilter` | `customerAdministration` > `grants` > `filter!` > `authenticationDomainId` |
| `MultiTenantAuthorizationGrantGroupIdInputFilter` | `customerAdministration` > `grants` > `filter!` > `groupId` |
| MultiTenantAuthorizationGrantScopeIdInputFilter | `customerAdministration` > `grants` > `filter!` > `scopeId` |
| MultiTenantAuthorizationRoleGroupIdInputFilter | `customerAdministration` > `roles` > `filter!` > `groupId` |
| MultiTenantIdentityUserGroupIdInput | `customerAdministration` > `users` > `filter!` > `groupId` | 
| MultiTenantIdentityUserIdInput | ~ | 
| OrganizationOrganizationCreateJobIdInput | `customerAdministration` > `jobs` > `organizationCreateAsyncResults` > `filter!` > `jobId` | 
| UserManagementGroupIdInput | ~ | 
| UserManagementUserIdInput | ~ |

PR #1134, opened 2024-05-07
fix(customeradministration): change auth_domain filter types in the package to use pointers

kporras07:
Currently they're not which means they are sent in the query and as a result the query always return 0 items even if they're there.

Note: I only need the first commit in this PR, but still pushed the rest of the tutone update.

### DIFF
--- a/.tutone.yml
+++ b/.tutone.yml
@@ -1599,6 +1599,15 @@ packages:
         field_type_override: "*OrganizationOrganizationCreateJobIdInput"
       - name: OrganizationOrganizationCreateJobStatusInput
         field_type_override: "*OrganizationOrganizationCreateJobStatusInput"
+        ######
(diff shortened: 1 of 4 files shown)


pranav-new-relic:
Thanks for the contribution again, @kporras07; welcome back :) I'll try and take a look soon in the next few days and get this done.

pranav-new-relic (approved):

pranav-new-relic:
Thanks again for the contribution, @kporras07. I've discarded tutone updates in other packages, just so the changes in this PR continue to remain around `customeradministration`.

I shall keep you posted on the timeline of having this fix released 🚀

pranav-new-relic:
@kporras07 changes from #1134 and #1140 should be out with v2.36.1 of the Go Client in some time.

PR #1135, opened 2024-05-07
Test/what happens if we generate everything again

sanderblue:
Just a draft to show what generating every all at once looks like right now.


sanderblue:
Closing since this was just for reference.

PR #1137, opened 2024-05-09
fix(changetracking): alter the functioning of milli and nanoseconds in timestamps

pranav-new-relic:
#1139 
#1151


pranav-new-relic:
Looks like changing the `serialization.EpochTime` datatype has major repercussions. We might need to manually change the timestamp going into the API call, only specific to ChangeTracking 🤔

pranav-new-relic:
will close this PR soon

PR #1138, opened 2024-05-10
fix(entity): Entity guid validation

NSSPKrishna:
This change will ensure that the customer-provided GUID is valid before it is accepted by the CLI, preventing any potential issues or errors that may arise from invalid input

### DIFF
--- a/pkg/entities/entity_guid_validator_test.go
+++ b/pkg/entities/entity_guid_validator_test.go
@@ -5,71 +5,70 @@ package entities
 
 import (
 	"encoding/base64"
+	"errors"
 	"fmt"
-	"strconv"
-	"strings"
+	"github.com/stretchr/testify/assert"
 	"testing"
 )
 
-func TestDecodeEntityGuid_Valid(t *testing.T) {
-	entity := DecodedEntity{
-		AccountId:  12345,
-		Domain:     "test_domain",
-		EntityType: "user",
-		DomainId:   "abc123",
-	}
-	encodedGuid := encodeEntity(entity)
-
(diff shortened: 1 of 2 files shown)


pkg/entities/entity_guid_validator_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+func TestDecodeEntityGuid_EmptyEntityType(t *testing.T) {
+	encodedGuid := base64.StdEncoding.EncodeToString([]byte("12345|domain||domainId"))
+
+	_, err := DecodeEntityGuid(encodedGuid)
+
+	if err != EntityGUIDValidationErrorTypes.EMPTY_ENTITY_TYPE_ERROR {
+		t.Errorf("Expected error 'entity type is required', got %v", err)
+	}
+}
```
Avinash-Sanpala, line 75: just a quick thought, I think it looks good if we use table driven tests here instead if multiple test cases

Avinash-Sanpala (dismissed):

pkg/entities/entity_guid_validator_test.go
NSSPKrishna: sure

pkg/entities/entity_guid_validator_test.go
Avinash-Sanpala, line 1: looks pretty clean!!

Avinash-Sanpala (approved):
LGTM!

GCamilleri:
@Avinash-Sanpala it looks like this has caused a bug in the change tracking cli command: https://github.com/newrelic/newrelic-cli/issues/1606

PR #1140, opened 2024-05-14
feat(organization): Add organizationUpdateSharedAccount mutation.

kporras07:
Hi!

We need this mutation for the project we're implementing this so adding it as PR.

Thanks!

_____

#### Maintainer's Note (for New Relic Internal Reviewers)
This change **does not** affect the New Relic Terraform Provider if consumed, since the `organization` package is not yet used there (as on 24 May, 2024).

### DIFF
--- a/.tutone.yml
+++ b/.tutone.yml
@@ -1455,8 +1455,8 @@ packages:
 #        max_query_field_depth: 2
 #      - name: organizationRevokeSharedAccount
 #        max_query_field_depth: 2
-#      - name: organizationUpdateSharedAccount
-#        max_query_field_depth: 2
+      - name: organizationUpdateSharedAccount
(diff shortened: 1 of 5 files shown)


pranav-new-relic:
Thanks for the contribution again, @kporras07. This seems like a pretty straightforward change. I shall take a final look at this by around the end of this week.

Can you also add a test case which can cover the new function you've added? This will help us make sure we have this function tested when relevant code changes are made. Thank you!

kporras07:
@pranav-new-relic thanks! Test added

pranav-new-relic (approved):

pranav-new-relic:
@kporras07 changes from #1134 and #1140 should be out with v2.36.1 of the Go Client in some time.

PR #1141, opened 2024-05-15
feat(alerts): add support for `titleTemplate` to NRQL alert conditions

baNROne:
* https://new-relic.atlassian.net/browse/NR-74508

* Adds support for `titleTemplate`
* Cleaned up some formatting while I was in there

* See this PR for related changes in the Terraform provider: https://github.com/newrelic/terraform-provider-newrelic/pull/2662

### Please do not merge yet


pkg/alerts/nrql_conditions.go
```diff
@@ -273,6 +275,7 @@ type NrqlConditionUpdateBase struct {
 	ViolationTimeLimitSeconds int                              `json:"violationTimeLimitSeconds,omitempty"`
 	Expiration                *AlertsNrqlConditionExpiration   `json:"expiration,omitempty"`
 	Signal                    *AlertsNrqlConditionUpdateSignal `json:"signal"`
+	IncidentTitleTemplate     *string                          `json:"incidentTitleTemplate,omitempty"`
```
akane0915, line 278: @baNROne I think we'll want to leave off `omitempty` for `NrqlConditionUpdateBase`. The reason is because we want a user to be able to remove an incident title template on an existing condition that already has the field configured. This presentation (slides 20-25 particularly) does a great job at explaining the reasoning. https://docs.google.com/presentation/d/1VEYfI7acg8LH_uGHjj5Su4RoF1lQRPVyfxK59xUMxV8/edit?usp=sharing

akane0915 (changes requested):
See inline comment

pkg/alerts/nrql_conditions.go
baNROne: Makes sense. Thanks @akane0915. I've updated it.

akane0915 (approved):
Looks good. Manually tested with the terraform PR.

akane0915 (approved):
Code looks good, but let's be sure to manually test once the changes are made in the backend and AGQL to update to `titleTemplate`

pranav-new-relic (approved):

PR #1143, opened 2024-05-22
fix(entity): Refactored to remove padding

NSSPKrishna:
base64.StdEncoding (default): Adds padding characters (=) during encoding for compatibility with standard base64. It removes padding during decoding.
base64.RawStdEncoding: Omits padding characters during encoding and expects the input string to not have any padding during decoding (throws error if present).


pranav-new-relic:
@NSSPKrishna please take a look at the lint failures

pranav-new-relic (approved):

PR #1146, opened 2024-05-29
feat(entities): add support for tags in EntityOutline types

abasha1234:
Add Support for Tags to the newrelic_entity Data Source

Ticket: https://new-relic.atlassian.net/browse/NR-273644


pkg/entities/entities_api.go
```diff
@@ -2079,36 +2079,64 @@ const getEntitySearchByQuery = `query(
 			permalink
 			reporting
 			type
+			tags {
+				key
+				values
+			}
```
pranav-new-relic, line 2085: we might not need this, can you please remove it (we can add it later when necessary)

pkg/entities/types.go
```diff
@@ -17345,6 +17345,7 @@ type EntityOutlineInterface interface {
 	GetGUID() common.EntityGUID
 	GetName() string
 	GetType() string
+	GetTags() []EntityTag
```
pranav-new-relic, line 17348: for anyone finding this PR in future: no other changes are needed in this file, since Tutone seems to have already populated all fields implementing the `EntityOutlineInterface` with a Tags attribute and a `GetTags` method.

This has been verified and checked multiple times by @abasha1234 and me.

PR #1171, opened 2024-06-05
chore(entities):  codegen cleanup

sanderblue:
Our `entities` package (along with a few others), contain a mix of manually edited code and generated code. The primary intent of this PR is to clean up and reorganize our `entities` package in a manner that allows us to generate code seamlessly for this package. 

This PR will require changes in [terraform-provider-newrelic](https://github.com/newrelic/terraform-provider-newrelic) and [newrelic-cli](https://github.com/newrelic/newrelic-cli) to accommodate updates to inputs and types. These changes should not be breaking changes.

Important design patterns to discuss as a team:
1. The file names of files that contain manually edited code now end with an underscore (`_`). E.g. `entities/entities_api_.go` and `entities/types_.go`. We have the option to flip this as well, i.e. make the generated files have the underscore, but that would require a massive overhaul with our code generation across the entire repo. 
2. Certain types that existed in the entities package, such as some Dashboard types, really ought to live in their respective packages. There will be caveats, but this will be improve scalability as we move towards maximum code generation.

<br>

**Associated PRs for dependent projects:**
- [Terraform provider](https://github.com/newrelic/terraform-provider-newrelic/pull/2697)
- [CLI](https://github.com/newrelic/newrelic-cli/pull/1623)


scripts/diff-types.js
```diff
@@ -0,0 +1,72 @@
...
+  const tutoneConfigFile = fs.readFileSync('.tutone.yml', 'utf8')
+  tutoneConfig = yaml.parse(tutoneConfigFile)
+
+  const schemaFileLatest = fs.readFileSync('schema.json', 'utf8');
+  schema = JSON.parse(schemaFileLatest);
+} catch(err) {
+  console.error(err);
+}
+
+const package = tutoneConfig.packages.find(x => x.name === 'entities');
```
NSSPKrishna, line 21: nice 👌

scripts/diff-types.js
```diff
@@ -0,0 +1,72 @@
...
+
+// console.log('packageTypes:', JSON.stringify(packageTypesByName, null, 2));
+// console.log('schemaTypes: ', JSON.stringify(schemaTypes, null, 2));
+// console.log('typesDiff:   ', typesDiff);
+// console.log('typesDiff:   ', typesDiff.length);
+
+
+const packageTypes = getPackageTypesFromTypesFile('/Users/sblue/dev/newrelic-client-go/pkg/entities/types.go');
+const packageTypesLatest = getPackageTypesFromTypesFile('/Users/sblue/Desktop/2024.06.04-entities-package-types-latest.json');
+
```
NSSPKrishna, line 40: These need to be updated to point to the correct path.

sanderblue: Oh right. We actually won't need this `diff-types.js` file. That was a script most to just gather information about what the differences were and why. I'll remove the file. Thanks for the reminder.

pkg/entities/entities_api.go
```diff
@@ -1320,28 +1592,6 @@ const getEntityQuery = `query(
...
-					corsUseNewrelicHeader
-					corsUseTracecontextHeaders
-					enabled
-					excludeNewrelicHeader
-				}
-				privacy {
-					cookiesEnabled
-				}
-			}
-		}
```
pranav-new-relic, line 1344: @sanderblue looks like Tutone's removing this portion, since the deepest (nested) fields of those at depth:2 are at a depth greater than 2. I just ran a `tutone generate --refetch --package entities` with 

```yml
          - name: entity
            max_query_field_depth: 3
            # changed 2 to 3 above
```

and this brought back this attribute `browserSettings`, with all of its nested attributes. While this might be what we'd want to do, this comes with the slight downside of modifying the entire query with additional fields, since the depth is now 3. 

What do you think we should do now - given that this block (unfortunately) is necessary and cannot be discarded (this is used by the Terraform resource `newrelic_browser_application`); do you think a change can be made with the query regenerated with depth 3? A point to also note is that the regeneration of this query with depth 3 doesn't seem to add any changes to `entities/types.go`, given it already seems to have types with their nested contents defined with the highest depth.

pkg/entities/entities_api.go
```diff
@@ -1890,14 +2308,16 @@ func (a *Entities) GetEntitySearchWithContext(
...
 	vars := map[string]interface{}{
-		"options":      options,
-		"query":        query,
-		"queryBuilder": queryBuilder,
-		"sortBy":       sortBy,
+		"options":             options,
+		"query":               query,
+		"queryBuilder":        queryBuilder,
+		"sortBy":              sortBy,
+		"sortByWithDirection": sortByWithDirection,
```
pranav-new-relic, line 2320: @sanderblue few thoughts - unfortunately, we weren't able to have `sortByWithDirection` so far, so by having this change to support the addition of `sortByWithDirection` move forward (which of course, is Tutone generated)
- we will be breaking `terraform-provider-newrelic` / `newrelic-cli` stuff using this function (though the fix would just be to add one tiny argument, so maybe it's not technically "breaking" :) )
- the same, for direct consumers of the New Relic Go Client - though again, the fix is just to add an extra empty argument

Do you think the above need to be considered? Maybe not entirely, in my opinion, since the only customer facing impact caused would be for direct consumers of this function from the Go Client, but I'd like to know your thoughts - if we do have to consider, we might just need to undo this change, and ignore when tutone shows a drift adding this argument everytime we run a `tutone generate` - but all of this is just what I think, your call; let me know what you think.

pkg/entities/entities_api_.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+		nextCursor
+	}
+	types {
+		count
+		domain
+		entityType
+		type
+	}
+} } }`
```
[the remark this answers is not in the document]
pranav-new-relic: I think it might also be useful to have a note added to `entities_api_.go` and `types_.go` which states something along the lines of "this code is NOT controlled by Tutone"

pranav-new-relic:
@sanderblue one more question 😄 in `entities/types.go`, looks like Tutone has been deleting a couple of interface unmarshal methods such as `func UnmarshalAiNotificationsAuthInterface()`, `func UnmarshalAiWorkflowsConfigurationInterface()`, `func UnmarshalApmApplicationEntityInterface()`, `func UnmarshalApmApplicationEntityOutlineInterface()` and changes around these functions (there are other references of deleted unmarshal functions too, but I believe they're all related to outdated types we have had for a very long time in the entities package).

Can you please take a look just so we can double check removing these methods is safe? I plan on taking a deeper look into `entities/types.go` tomorrow when I'm back again, but just wanted to know your opinion about this.

pranav-new-relic:
(_edited_ version of the comment posted on 19 Jun 2024)

@sanderblue one more question 😄 in `entities/types.go`, looks like Tutone has been deleting a couple of interface unmarshal methods such as  `func UnmarshalApmApplicationEntityInterface()`, `func UnmarshalApmApplicationEntityOutlineInterface()` and changes around these functions (there are other references of deleted unmarshal functions `UnmarshalJSON` too, along with other things like `func UnmarshalDashboardWidgetCommonsInterface()`, `func UnmarshalEntitySummaryMetricValueInterface()` and similar ones but I believe these are mostly related to outdated types we have had for a very long time in the entities package).

Can you please take a look just so we can double check removing these methods is safe? I plan on taking a deeper look into `entities/types.go` tomorrow when I'm back again, but just wanted to know your opinion about this.

I see `func UnmarshalAiWorkflowsConfigurationInterface()` and `func UnmarshalAiNotificationsAuthInterface()` have also been removed from `types.go`, but **moved** to `types_.go`, which is why I'm wondering the ones I've listed above also need to be in `types_.go` maybe? Just a thought 🤔

pkg/entities/entities_api_.go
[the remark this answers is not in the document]
sanderblue: Pretty sure this query supports tags for all types with the `tags` field at the top level of the response fields (lines 72-75).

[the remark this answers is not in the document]
pranav-new-relic: I get it now - this should be fine then :) except for one final find - the field `type` seems to have been excluded in this query.

https://www.diffchecker.com/2EpPBVXv/

Also, I apologize for the silly question, but just trying to be 100% sure - can we be sure that the addition of `tags` in the list of attribute common to all EntityOutline types would suffice - and we wouldn't need to add them per type?

pkg/entities/entities_api.go
sanderblue: Interesting, I'm surprised this didn't break tests, but you're right we can't remove this. I'll update with your suggestion and see how things go.

pkg/entities/entities_api.go
sanderblue: While this requires a change in the provider as well as in the CLI, this technically isn't a breaking change for those projects. I have a [PR](https://github.com/newrelic/terraform-provider-newrelic/pull/2697) to update the provider to support this change 🙂

pranav-new-relic: Yeah, agreed - which is why I mentioned in my comment too that this isn't supposed to be a "breaking" change :) my only concern is for those who could be using this function from the Go Client directly in other Go-based stuff, but I think this should be okay 👍 thanks for taking a look

pkg/entities/entities_api_.go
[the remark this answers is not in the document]
sanderblue: Yes, the `tags` field does get applied to the `ApmApplicationEntityOutline` type in the given GraphQL query. In GraphQL, when you query for a field that is part of an interface or a union, the fields within the fragment of a specific type will include the fields from the parent type or interface unless explicitly overridden.

Also the `type` field is on line 76 😉  It just moved during the code generation (my hunch is the iteration that's involved in codegen does not guarantee the order of the fields)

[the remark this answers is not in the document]
pranav-new-relic: how did I not see this - I'm so very sorry about that. I think we should be good with resolving this comment - thank you so much for explaining :)

sanderblue:
> (_edited_ version of the comment posted on 19 Jun 2024)
> 
> @sanderblue one more question 😄 in `entities/types.go`, looks like Tutone has been deleting a couple of interface unmarshal methods such as `func UnmarshalApmApplicationEntityInterface()`, `func UnmarshalApmApplicationEntityOutlineInterface()` and changes around these functions (there are other references of deleted unmarshal functions `UnmarshalJSON` too, along with other things like `func UnmarshalDashboardWidgetCommonsInterface()`, `func UnmarshalEntitySummaryMetricValueInterface()` and similar ones but I believe these are mostly related to outdated types we have had for a very long time in the entities package).
> 
> Can you please take a look just so we can double check removing these methods is safe? I plan on taking a deeper look into `entities/types.go` tomorrow when I'm back again, but just wanted to know your opinion about this.
> 
> I see `func UnmarshalAiWorkflowsConfigurationInterface()` and `func UnmarshalAiNotificationsAuthInterface()` have also been removed from `types.go`, but **moved** to `types_.go`, which is why I'm wondering the ones I've listed above also need to be in `types_.go` maybe? Just a thought 🤔

- I'm not sure why `UnmarshalApmApplicationEntityInterface()` was deleted, but my hunch is the interface definition what it implements changed to where Tutone no longer needs to create a "special" unmarshal method for it. And if I add back the [function](https://github.com/newrelic/newrelic-client-go/blob/main/pkg/entities/types.go#L16744-L16787) that's currently in `main`, that unmarshal method references a [type](https://github.com/newrelic/newrelic-client-go/blob/main/pkg/entities/types.go#L16767C9-L16767C49) that doesn't exist anymore (`ApmAgentInstrumentedServiceEntityOutline`). And since that type was the only type referenced in the "special" unmarshal method, I think that's why we no longer need that unmarshal method now. 

- The `UnmarshalJSON()` still exists in several areas where it's required for special unmarshalling, but I'm fairly certain type definitions have changed so some no longer need this method I think. 

- The type `EntitySummaryMetricValue` no longer exists (see screenshot), so we're good there.
Screenshot 2024-06-26 at 12 46 58 PM

- The type `DashboardWidgetCommons` no longer exists as well, so I think we're also good there.

- The `AiNotificationsAuth` and `AiWorkflowsConfiguration` types and associated methods were manually added a while back when we had to do a quick fix for a breaking change that was introduced in the API. We will want to put these types in their proper package at some point, but I didn't want to take any additional risks on top what's already being performing this massive regeneration of code.

pranav-new-relic (dismissed):

pranav-new-relic:
Thanks for clarifying - I guess we're all set :)

noahmmcgivern (approved):

PR #1173, opened 2024-06-11
fix(customeradministration): TargetId should be nullable.

kporras07:
Otherwise it gets sent in the query and we get no results even if there are expected results


pranav-new-relic:
@kporras07 👍 will quickly perform one final check to make sure no unexpected tests are breaking, and will go ahead and merge this.

pranav-new-relic (approved):

PR #1177, opened 2024-06-23
Add update original message to destination configurations

nzur-newrelic:


pkg/workflows/workflows_integration_test.go
```diff
@@ -30,7 +30,7 @@ func TestIntegrationCreateWorkflow(t *testing.T) {
 	notificationTriggers := []AiWorkflowsNotificationTrigger{"ACTIVATED"}
 
 	// Create a workflow to work with in this test
-	workflowInput := generateCreateWorkflowInput(channel, notificationTriggers)
+	workflowInput := generateCreateWorkflowInput(channel, notificationTriggers, nil)
```
nzur-newrelic, line 33: Sadly there's no real way to check the flag as it's only relevant for SLACK.
For the moment I'm just checking the fact that it's nil, but in the future we will expand its usage further so more tests can be done.

pranav-new-relic (approved):

pranav-new-relic:
image

PR #1182, opened 2024-06-28
chore(entities): manual adjustments were needed for some dashboard types (pointers)

sanderblue:


pranav-new-relic:
@sanderblue just thinking aloud to make sure I'm on the right path with understanding the necessity of these changes - please let me know if I'm right about the following.

- Prior to our entities package cleanup PR merge and release, this is how the `DashboardVariable` struct looked like in the entities package. (the following block is from `v2.37.1` of the client, released before the `entities` package cleanup PR was merged)
https://github.com/newrelic/newrelic-client-go/blob/8a5c6dc32520fda24a39ad120e474c91edad2bbe/pkg/entities/types.go#L7975-L7997
- However, in the cleanup procedure, when we ran a `tutone generate --refetch --package entities`, this led to the removal of pointers of most datatypes in DashboardVariable, since these most probably could have been manual changes made out of the scope of Tutone in the past (I'm guessing) (the following block is from `v.2.38.0` of the client, released after the `entities` package cleanup PR was merged)
https://github.com/newrelic/newrelic-client-go/blob/bd0bfe8af3e7aff3447747bd62022bf629aaea32/pkg/entities/types.go#L5275-L5297
- In order to restore these changes, you're trying to bring this back by adding necessary changes to `.tutone.yml`, and also moving this to `types_.go` in the `entities` package so as to also keep it out of Tutone scope for now.

pkg/entities/types.go
```diff
@@ -5272,34 +5272,10 @@ type DashboardTableWidgetConfiguration struct {
...
-	Title string `json:"title,omitempty"`
-	// Specifies the data type of the variable and where its possible values may come from.
-	Type DashboardVariableType `json:"type,omitempty"`
-}
-
 // DashboardVariableDefaultItem - Represents a possible default value item.
 type DashboardVariableDefaultItem struct {
 	// The value of this default item.
-	Value DashboardVariableDefaultValue `json:"value,omitempty"`
+	Value *DashboardVariableDefaultValue `json:"value,omitempty"`
```
pranav-new-relic, line 5278: @sanderblue if my understanding I've shared in the previous comment ⬆️ is true, I think all of these changes are fine 👍 I've compared these with the changes in 2.37.1 of the client and they are entirely in sync. 

However, I see the section highlighted in **this** comment, `*DashboardVariableDefaultValue` inside `DashboardVariableDefaultItem` as the only exception (the only change that does not exist in versions of the client prior to the entities cleanup PR merge). 

Now that `DashboardVariableDefaultValue` is under Tutone scope (is being generated as `*DashboardVariableDefaultValue`), and since DashboardVariableDefaultItem contains `DashboardVariableDefaultValue`, this is also being changed to `*DashboardVariableDefaultValue`. 

Unfortunately, this was not the case with in versions of the client before the `entities` cleanup PR was merged. It looks like a manual change was made to `*DashboardVariableDefaultValue` and `*[]DashboardVariableDefaultItem` in `DashboardVariable` to be taken as a pointer, but because the change to  `*DashboardVariableDefaultValue` was not Tutone driven, this did not propagate into `DashboardVariableDefaultValue` inside `DashboardVariableDefaultItem`, as a result of which this is what `DashboardVariableDefaultItem` looks like in v2.37.1 of the client.

https://github.com/newrelic/newrelic-client-go/blob/8a5c6dc32520fda24a39ad120e474c91edad2bbe/pkg/entities/types.go#L7976-L7980

https://github.com/newrelic/newrelic-client-go/blob/8a5c6dc32520fda24a39ad120e474c91edad2bbe/pkg/entities/types.go#L8005-L8009

This should absolutely be the right change, but I believe with this we might need additional changes in the Terraform Provider, so just wanted to post this as a thought.

sanderblue: I think we're good to go for this PR. Overall, what you said is correct with regard to how I made these changes with our tooling. I will create a followup PR if anything else comes up in the provider.

noahmmcgivern (approved):

PR #1184, opened 2024-07-01
feat(nrql_alert_conditions): Add data_account_id for NRQL alert condition

founddrama:
Adds support for the new `data_account_id` field for NRQL alert conditions. This field already appears as `dataAccountId` in the `nrql` block for NRQL alert conditions in the NerdGraph GraphQL API. This new field is entering a limited preview and is feature flagged such that it has customer visibility for read operations, but creates and updates are guarded with the feature flag in the downstream service.

This field is a prerequisite for support being added in the [Terraform provider](https://github.com/newrelic/terraform-provider-newrelic) (_pull request pending_)

PR #1185, opened 2024-07-02
chore(organization): fix errors caused by change in API error handling

pranav-new-relic:


pranav-new-relic:
image

the first is an intermittent failure,
image

the second is expected, has been reported to the team (a probable consequence of #1184)

should be good to go 👍

PR #1198, opened 2024-07-31
feat(synthetics): Added support for Synthetics Multi Browser, Multi Devices attributes in few synthetics APIs

shashank-reddy-nr:
### Background
Recently, Synthetics added a new feature for scripted, step, and simple browser monitors to support multiple browsers and devices, allowing users to choose from various options when running monitors. Until now, we only had the option of choosing one device combination. As per customer requests and Synthetics' requirements to support the same multi-browser/devices functionality in Terraform, So, support terraform changes we need changes in client-go as well. So, made below changes to support this.

### Changes
1: Ran tutone and generated synthetics package, which brought browsers, devices fileds.
2: Added integration test cases to validate browsers, devices fields.
3: Sross checked the changes tutone brought in and tested the changes related to these fields along with integration test cases as well.

### DIFF
--- a/pkg/synthetics/synthetics_api.go
+++ b/pkg/synthetics/synthetics_api.go
@@ -582,7 +582,9 @@ const SyntheticsCreateScriptBrowserMonitorMutation = `mutation(
 		advancedOptions {
 			enableScreenshotOnFailureAndScript
 		}
+		browsers
 		createdAt
+		devices
 		guid
 		id
 		locations {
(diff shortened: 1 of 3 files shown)


pranav-new-relic:
image

LGTM

pranav-new-relic (approved):

pranav-new-relic:
we might need to make a *[] later on probably, when the API supports accepting an empty list (to remove browsers and devices)

shashank-reddy-nr:
Agree, we will do this whenever required

PR #1212, opened 2024-09-24
feat(codegen): generate code based on latest API changes

nr-developer-toolkit:
Update generated code using Tutone


sanderblue:
Getting closer :)

PR #1218, opened 2024-09-25
feat(automation): generate code based on latest API changes

nr-developer-toolkit:
Update generated code using Tutone


sanderblue:
Note: It looks like the calculation for `max_query_field_depth` still needs some adjustment because this should be removing some of the code in this PR. Will continue to fine tune this.

sanderblue:
Note: It looks like the calculation for `max_query_field_depth` still needs some adjustment because this should be removing some of the code in this PR. Will continue to fine tune this.

PR #1238, opened 2024-10-11
Fix/codeql action

vaughnw128:
Fixing the CodeQL action to ignore non-required (testing) files using `codeql-config.yml`. Checkout HEAD^2 has also been removed as it is no longer required and generates warnings. Errors are generated in the security tab if significant quantities of files are not scanned by CodeQL, so this should produce good findings.

image

image


pranav-new-relic:
Thanks for this, @vaughnw128

pranav-new-relic (approved):

PR #1255, opened 2024-11-06
fix(entities): remove relationships field from entity, entities, entitySearch queries

pranav-new-relic:
replaces #1249


vagrawal-newrelic (approved):
approved

PR #1266, opened 2025-01-29
feat(workflows): added new notification trigger type called investigating

mvaishya:
# Description
Jira Ticket: NR-348444.

Added Investigate trigger state, which performs similar function as acknowledge state.

# Testing
Unit or Integration testing is not required for this change as no explicit tests are written for other triggers


pranav-new-relic:
covered via #1270

PR #1267, opened 2025-01-30
feat(application_settings): amp application settings to sync with the latest released/available features/functionalities in apm settings ui

vinay-newrelic:
Right now, this TF resource only supports a subset of the application settings available on the APM UI .
This PR concludes the changes required for `newrelic_application_settings` resource to add the missing set of configuration settings that are available on the UI


.tutone.yml
```diff
@@ -1675,4 +1675,3 @@ packages:
       - name: EpochMilliseconds
         field_type_override: "*nrtime.EpochMilliseconds"
         skip_type_create: true
-
```
pranav-new-relic, line 1678: @vinay-newrelic please revert this change and also everything else made for spellchecks in other packages

pkg/agentapplications/types.go
```diff
@@ -295,7 +297,7 @@ type AgentApplicationSettingsApmConfigInput struct {
 	// The desired target for the APDEX measurement of this application.
 	ApdexTarget float64 `json:"apdexTarget,omitempty"`
 	// Sets if installed agents should override local settings with ones set here.
-	UseServerSideConfig bool `json:"useServerSideConfig,omitempty"`
+	UseServerSideConfig *bool `json:"useServerSideConfig,omitempty"`
```
pranav-new-relic, line 300: @vinay-newrelic for this and **all** similar changes below (`bool` -> `*bool`) can you please add a comment right above the line which might look like

https://github.com/newrelic/newrelic-client-go/blob/cbb232b562700205ed14d7fdb9786abbf9cfb024/pkg/dashboards/types.go#L293

this would make sure no one fiddles with these with a tutone command run; they'd be able to see this comment when changes made by Tutone are reviewed

pkg/agentapplications/types.go
```diff
@@ -629,7 +631,7 @@ type AgentApplicationSettingsUpdateError struct {
 // [Ruby agent configuration](https://docs.newrelic.com/docs/agents/ruby-agent/configuration/ruby-agent-configuration/)
 type AgentApplicationSettingsUpdateInput struct {
 	// The new name for the application.
-	Alias string `json:"alias,omitempty"`
+	Alias *string `json:"alias,omitempty"`
```
pranav-new-relic, line 634: @vinay-newrelic just curious; why a `*` with a `string` 🤔

vinay-newrelic: Yeah, it is not longer needed as i just wanted to initially identify an empty string with nil value but finally a simple string was sufficient in our case. Will revert this change.

pranav-new-relic:
@vinay-newrelic all good here, except for one question (just adding it here for the record) - I think we'd discussed sometime that you intend to move the mutation to update application settings into a different package to avoid type conflicts, etc. Do you think that is still needed, or are we good with the current arrangement? I don't think there's anything incorrect, but would just like to double check

vinay-newrelic:
> @vinay-newrelic all good here, except for one question (just adding it here for the record) - I think we'd discussed sometime that you intend to move the mutation to update application settings into a new/different package to avoid type conflicts, etc. Do you think that is still needed, or are we good with the current arrangement? I don't think there's anything incorrect, but would just like to double check

@pranav-new-relic Just thought of moving it to seperate package since the current apm package is a but messy and overloaded but then realised agent applications pkg is the one which already has this schema and mutations which is also pretty organised.shifting it under APM package again would result in redundancy and introduce confusion.

nr-developer-toolkit (approved):

pranav-new-relic (approved):

pranav-new-relic:
All good 👍 we're good to go

PR #1270, opened 2025-01-31
feat(workflows): adds new `notification_trigger` type `INVESTIGATING`

shubh-0708:
Description
Jira Ticket: NR-348444.

Added Investigate trigger state, which performs similar function as acknowledge state.

Testing
Unit or Integration testing is not required for this change as no explicit tests are written for other triggers


pranav-new-relic (approved):

pranav-new-relic:
Please do not merge right now

PR #1278, opened 2025-03-27
fix(build): added codecov token to be used from secrets for reporting code coverage

vinay-newrelic:
Addition of codecov token


vagrawal-newrelic (approved):
LGTM

PR #1297, opened 2025-05-06
feat: adds entityrelationship pkg

vinay-newrelic:
Adds a new entity relationship package


pkg/entityrelationship/entityrelationship.go
```diff
@@ -0,0 +1,33 @@
...
+	config config.Config
+}
+
+// nolint
+func newMockResponse(t *testing.T, mockJSONResponse string, statusCode int) Entityrelationship {
+	ts := mock.NewMockServer(t, mockJSONResponse, statusCode)
+	tc := mock.NewTestConfig(t, ts)
+
+	return New(tc)
+}
```
pranav-new-relic, line 23: moving this to the test file should most probably eliminate the necessity of `//nolint` - can you please check with other packages to see if this is possible, and correct this (e.g. `usermanagement`)

pkg/entityrelationship/entityrelationship_integration_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"testing"
+
+	"github.com/newrelic/newrelic-client-go/v2/pkg/common"
+	"github.com/newrelic/newrelic-client-go/v2/pkg/testhelpers"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	sourceEntityGuidVal = "MzgwNjUyNnxFWFR8U0VSVklDRV9MRVZFTHw1ODA4MDM"
+	targetEntityGuidVal = "MzgwNjUyNnxFWFR8U0VSVklDRV9MRVZFTHw1NzE0Nzk"
```
pranav-new-relic, line 16: can you move these to the `testhelpers` package where you'll find a bunch of other GUIDs too we've saved into variables to reuse across tests in packages

pkg/entityrelationship/entityrelationship_integration_test.go
```diff
@@ -0,0 +1,69 @@
...
+	createResult, err := client.EntityRelationshipUserDefinedCreateOrReplace(
+		sourceEntityGUID,
+		targetEntityGUID,
+		relationshipType,
+	)
+
+	require.NoError(t, err)
+	require.NotNil(t, createResult)
+	require.Empty(t, createResult.Errors)
+
```
pranav-new-relic, line 43: looks like we're doing a create and then a delete directly
- are we not doing an update because it would be the same function to use anyway `EntityRelationshipUserDefinedCreateOrReplace`? If so, I think we should still add a duplicate so the integration test finds faults with the API if the mutation (somehow) works with create and not update
- can you also add a check for read here from the `entities` package to validate that the added entity relationship actually exists (so we cover CRUD)

vinay-newrelic: Yeah that was reason i did not throw a light on update function but makes sense to include update and read as u suggested.So added the required changes

PR #1305, opened 2025-06-11
feat: add BillboardSettings & ThresholdWithSeriesOverrides to Dashboard

jcrengifoa:


pkg/dashboards/dashboards_types.go
```diff
@@ -395,3 +397,70 @@ type DashboardWidgetColorOverrides struct {
...
+
+type ThresholdSeverity string
+
+const (
+	ThresholdSeveritySuccess     ThresholdSeverity = "success"
+	ThresholdSeverityWarning     ThresholdSeverity = "warning"
+	ThresholdSeveritySevere      ThresholdSeverity = "severe"
+	ThresholdSeverityCritical    ThresholdSeverity = "critical"
+	ThresholdSeverityUnavailable ThresholdSeverity = "unavailable"
+)
```
shashank-reddy-nr, line 455: Our general practice is to keep any constants at the top of the file. We recommend using capital letters for constant strings, though you can use lowercase if your API requires it.

jcrengifoa: Got it @shashank-reddy-nr, the variables have been moved + changed const in favor of var to further comply with the code standards

PR #1312, opened 2025-06-18
feat(nrdb): add support to the nrdb package to support facet, timeseries result unmarshaling 

pranav-new-relic:


shashank-reddy-nr (approved):
LGTM

PR #1315, opened 2025-06-25
fix(alerts): fix incorrect delete endpoint for multilocation alert conditions

pranav-new-relic:
Changes to be merged to `main` from
- #1314


nr-developer-toolkit (approved):

pranav-new-relic:
The outcome of a very insightful conversation with: @jstorerNR - we'll move forward with this PR once he has the change to handle traffic to the older `alerts_conditions` DELETE endpoint (caused by older versions of the client consumed by older versions of the New Relic Terraform Provider) worked around/handled in the new service they're moving to

PR #1323, opened 2025-07-28
fix(change-tracking): remove test print

Nandu-pns:
This update simply removes a test printout that currently appears when you execute any change tracking CLI command.

image

Apart from above test print changes, this PR also addresses the test failures with below errors.
https://github.com/newrelic/newrelic-client-go/actions/runs/16562626029/job/46835246933

- Fixes the tests failing with `strconv.ParseInt: parsing 'null': invalid syntax`
- EpochTime.UnmarshalJSON method in epoch_time.go couldn't handle null values from the API
- changetracking.Seconds type couldn't unmarshal numeric values from API responses
- ServiceLevelIndicator.ID field was defined as int but API was returning ID values as strings


NSSPKrishna (dismissed):

Nandu-pns:
Change Tracking tests are now passing. The other integration test failures are unrelated to this PR and appear to be intermittent.

image

PR #1325, opened 2025-08-07
fix(change-tracking): update depth to 2 and fix test failures

Nandu-pns:
https://new-relic.atlassian.net/browse/NR-448202


pranav-new-relic (approved):

pranav-new-relic:
@Nandu-pns looks good to me - but before we merge, can you please ensure that the changes generated in `changetracking > types.go` upon running `tutone generate --refetch --package changetracking` are unnecessary (and are not being missed out on, for anything we might need right now)? Some of them only seem like semantic differences, but I wasn't able to go through all of the changes :) 

Can you please check - and then I think we should be good

pranav-new-relic (approved):

PR #1401, opened 2026-04-28
fix(alerts): add policy entity guid

gffuentes:
This adds entityGuid field in the response for policies for the nerdgraph methods. This field was recently added and has already been stitched. After this is merged in, I will add a PR to the terraform provider to give terraform policy resource the entityGuid field, allowing users to leverage tag creation via the policy.entityGuid.

jira: https://new-relic.atlassian.net/browse/NR-529749

### DIFF
--- a/pkg/alerts/policies.go
+++ b/pkg/alerts/policies.go
@@ -363,6 +363,7 @@ const (
 						name
 						incidentPreference
 						accountId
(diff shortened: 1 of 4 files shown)


pranav-new-relic (approved):
LGTM

PR #1403, opened 2026-05-05
fix(fleetcontrol): restore managedEntityType in AgentConfigurationEntity GetEntity fragments

pranav-new-relic:
## Summary

- Adds `managedEntityType` back to the `EntityManagementAgentConfigurationEntity` inline fragment in both `getEntityQuery` and `getEntitySearchQuery`
- It was removed in #1402 as a precaution alongside the real nullability fix (custom `UnmarshalJSON` on `EntityManagementActorStitchedFields`)
- With the unmarshal now robust, the field is safe to re-add for this specific entity type
- Required by [terraform-provider-newrelic#3079](https://github.com/newrelic/terraform-provider-newrelic/pull/3079) so the import flow can populate `managed_entity_type` from the API response

## Test plan

- [ ] Local build passes: `go build ./pkg/fleetcontrol/...`
- [ ] Integration tests in terraform-provider-newrelic pass with this client version

### DIFF
--- a/pkg/fleetcontrol/fleetcontrol_api.go
+++ b/pkg/fleetcontrol/fleetcontrol_api.go
@@ -636,6 +636,7 @@ const getEntityQuery = `query(
 		__typename
 		agentType
 		configurationType
+		managedEntityType
 		metadata {
 			createdAt
 			createdBy {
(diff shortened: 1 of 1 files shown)


pranav-new-relic:
Closing — this approach doesn't work. Adding managedEntityType to the AgentConfigurationEntity fragment causes a GraphQL schema validation error: 'fields have different nullability shapes'. The field has conflicting nullability across entity types in the union, so the server rejects the query entirely. The terraform-provider-newrelic import will use a composite ID instead to carry managed_entity_type through import.

PR #1417, opened 2026-05-22
test: fix flaky integration tests across multiple packages

pranav-new-relic:
## Summary

Fixes several integration test failures observed in CI that pass locally but fail under concurrent load or after entity creation:

- **fleetcontrol**: Switch to `NEW_RELIC_FLEET_TEST_*` credentials (isolated account for fleet resources); rewrite tests into three self-contained workflows (fleet lifecycle, configuration lifecycle, deployment/managed-entity read-only search); remove all hardcoded GUIDs; add package-level comment explaining why managed-entity and deployment write paths are excluded; fix defer cleanup registration order so resources are never leaked on mid-test assertion failures
- **keytransaction**: Add 5-second sleep after create before update — `KeyTransactionUpdate` resolves `serviceLevel.indicators` on the newly created entity, which NerdGraph cannot resolve until the entity is indexed; without the sleep the call exhausts all 3 retries under CI load
- **servicelevel**: `GetIndicators` was called with the SLI's own GUID instead of the parent application entity GUID; `serviceLevel { indicators }` only resolves on the owning entity, so the SLI GUID always returns nothing; also add defer-based best-effort cleanup so SLIs are deleted even on mid-test failures
- **usermanagement**: `UserManagementGroupCleanupForIntegrationTests` propagated "Could not find the target or you are unauthorized" from `UserManagementDeleteGroup`, causing `TestIntegrationGroupManagementWithUsers` to fail when parallel tests race to delete the same leftover group
- **entityrelationship**: Fix type assertion (`ExternalEntity` → `MobileApplicationEntity`) and refresh stale test entity GUIDs to active Mobile Application entities
- **CI (GitHub Actions)**: Wire `NEW_RELIC_FLEET_TEST_*` secrets into the integration test job

## Test plan

- [ ] CI integration tests pass for all affected packages: `fleetcontrol`, `keytransaction`, `servicelevel`, `usermanagement`, `entityrelationship`
- [ ] Fleet tests skip gracefully when `NEW_RELIC_FLEET_TEST_API_KEY` is not set
- [ ] No dangling resources left after a mid-test assertion failure (defer guard pattern verified)


nr-developer-toolkit (dismissed):

nr-developer-toolkit (approved):
LGTM

PR #1421, opened 2026-05-29
feat: add DisableEventCreation field to NrqlConditionTerm

tabathadelane:
## What

Adds `DisableEventCreation *bool` to `NrqlConditionTerm` behind the `DCON/disable_event_creation_apis` feature flag (gated at the GraphQL layer).

## Changes

[IDD](https://newrelic.atlassian.net/wiki/x/7YDHMwE)

- `nrql_conditions.go` - `DisableEventCreation *bool` on `NrqlConditionTerm` struct + `disableEventCreation` added to the GraphQL query terms block
- `nrql_conditions_integration_test.go` - `TestIntegrationNrqlConditions_DisableEventCreation` with create/get/update/nil test cases

## Jira

[NR-566759](https://new-relic.atlassian.net/browse/NR-566759)

## Dependencies

Requires [alerts_graphql_service#729](https://source.datanerd.us/Alerting/alerts_graphql_service/pull/729) to be merged and deployed first.

[NR-566759]: https://new-relic.atlassian.net/browse/NR-566759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1425, opened 2026-06-08
test(federated-logs): skip unit and integration tests blocking CI

pranav-new-relic:
## What

Unblocks CI on PRs that touch unrelated packages (e.g. #1421) by skipping the federated-logs tests that are currently failing for reasons orthogonal to the changes under review.

## Changes

- **`federatedlogs_unit_test.go`** — switched its build tag from `unit` to a placeholder (`unit_fedlogs_skip`). With `-tags unit`, the file currently fails to compile because the API signatures for `FederatedLogs{Create,Update}{Setup,Partition}` and `Get{Setup,Partition}` now take a leading `accountID int`, and the test calls have not been updated. Disabling the file is the minimal unblock; the proper fix (updating each call site) will land in a follow-up PR.
- **`federatedlogs_integration_test.go`** — added `t.Skip(...)` at the top of `TestIntegrationFederatedLogs_Setup` and `TestIntegrationFederatedLogs_Partition`. Both fail in CI with `Exception while fetching data (/createSetup) : User is not authorized to perform this action`, indicating the fleet-test account does not currently have `federatedLogsCreateSetup` access. Skip until access is granted.

`TestIntegrationFederatedLogs_AwsConnection` still runs.

## Why

Both failures are unrelated to in-flight feature work but block the shared `test-unit` and `test-integration` jobs on every PR.

## Test plan

- [x] `go test -count=1 -tags=unit ./pkg/federatedlogs/...` → ok
- [x] `go test -count=1 -tags=integration -run 'TestIntegrationFederatedLogs_(Setup|Partition)$' -v ./pkg/federatedlogs/...` → both report `--- SKIP` with the access reason
- [ ] CI green on this PR

## Follow-up

Separate PR will re-enable the unit tests by passing `accountID` through to the six call sites, and re-enable the integration tests once the fleet-test account is granted access.

### DIFF
--- a/pkg/federatedlogs/federatedlogs_integration_test.go
+++ b/pkg/federatedlogs/federatedlogs_integration_test.go
@@ -71,6 +71,7 @@ func TestIntegrationFederatedLogs_AwsConnection(t *testing.T) {
 }
 
 func TestIntegrationFederatedLogs_Setup(t *testing.T) {
+	t.Skip("skipping until the fleet-test account is granted federatedLogsCreateSetup access; currently fails with 'User is not authorized to perform this action'")
 	t.Parallel()
 
 	testAccountID, err := mock.GetFleetTestAccountID()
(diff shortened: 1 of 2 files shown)


nr-developer-toolkit (approved):

pranav-new-relic:
This PR had to be reverted as the tests seem to be breaking execution in a lot of PRs, disallowing verification. After we're sure the test user has the right capabilities, these tests shall be restored.

PR #1429, opened 2026-07-08
feat(entities): add cursor pagination for entity search

pranav-new-relic:
EntitySearch responses include a nextCursor but the existing methods had no way to feed it back in, so callers could only ever see the first page. Adds GetEntitySearchByQueryWithCursor and GetEntitySearchWithCursor (with WithContext variants) that accept a cursor argument alongside the existing parameters; the derived query strings inject $cursor: String and forward it into results(cursor: $cursor).

### DIFF
--- a/pkg/entities/entities_api_.go
+++ b/pkg/entities/entities_api_.go
@@ -2,6 +2,8 @@ package entities
 
 import (
 	"context"
+	"fmt"
+	"strings"
 )
 
 // Search for entities using a custom query.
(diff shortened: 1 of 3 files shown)


pranav-new-relic:
## Trying this out before merge

If you'd like to consume this branch to test the cursor-pagination fix, you can point your `go.mod` at this commit directly. Two ways:

### Option 1: `go get` at the branch commit (simplest)

From your project root:

```bash
go get github.com/newrelic/newrelic-client-go/v2@c8a6dfcc
```

That pins to the exact commit on this branch and updates `go.mod` / `go.sum` with a pseudo-version like `v2.90.1-0.20260708...c8a6dfcc`. `go get github.com/newrelic/newrelic-client-go/v2@feat/entity-search-cursor` also works and will re-resolve to whatever the latest commit on the branch is.

### Option 2: `replace` directive against a local checkout

Useful if you want to iterate against a local clone:

```bash
git clone --branch feat/entity-search-cursor https://github.com/newrelic/newrelic-client-go.git ../newrelic-client-go
```

Then in your project's `go.mod`:

```gomod
replace github.com/newrelic/newrelic-client-go/v2 => ../newrelic-client-go
```

Run `go mod tidy` and you're pointed at the checkout.

### Using the new functions

Two new entry points, mirroring the existing pair — pass an empty string for the first page, then feed `Results.NextCursor` back in until it's empty:

```go
package main

import (
	"context"
	"fmt"

	"github.com/newrelic/newrelic-client-go/v2/newrelic"
	"github.com/newrelic/newrelic-client-go/v2/pkg/config"
	"github.com/newrelic/newrelic-client-go/v2/pkg/entities"
)

func main() {
	client, err := newrelic.New(newrelic.ConfigPersonalAPIKey("NRAK-..."))
	if err != nil {
		panic(err)
	}

	cursor := ""
	page := 0
	for {
		page++
		res, err := client.Entities.GetEntitySearchByQueryWithCursorWithContext(
			context.Background(),
			entities.EntitySearchOptions{},
			"domain = 'APM' AND reporting = 'true'",
			nil,     // sortBy
			cursor,  // "" on the first call
		)
		if err != nil {
			panic(err)
		}

		fmt.Printf("page %d: %d entities\n", page, len(res.Results.Entities))
		for _, e := range res.Results.Entities {
			fmt.Printf("  %s  %s\n", e.GetGUID(), e.GetName())
		}

		if res.Results.NextCursor == "" {
			break // no more pages
		}
		cursor = res.Results.NextCursor
	}
}
```

The `queryBuilder` variant is identical in shape — call `GetEntitySearchWithCursor(options, "", queryBuilder, sortBy, sortByWithDirection, cursor)` and drive the same loop.

Config knobs (unchanged from the existing methods):
- `config.Region` / `newrelic.ConfigRegion` if you're on EU
- `EntitySearchOptions.TagFilter` if you want extra tag columns in the response

Happy to answer any questions if something doesn't compile on your side.

PR #1434, opened 2026-07-16
feat(region): add GOV/FedRAMP region with compliant endpoints

pranav-new-relic:
## Summary

Adds `GOV` as a new region (also parseable as `fedramp`) with FedRAMP-compliant endpoint URLs, following the same pattern as the existing `JP` region.

### Endpoints added

| Field | GOV URL | Notes |
|---|---|---|
| NerdGraph | `https://gov-api.newrelic.com/graphql` | Confirmed 200 with GOV credentials |
| REST | `https://gov-api.newrelic.com/v2` | Same host as NerdGraph |
| Infrastructure | `https://gov-infra-api.newrelic.com/v2` | Per FedRAMP docs; confirmed 401→200 with `fedramp: true` on infra agent |
| Insights (Event API) | `https://gov-insights-collector.newrelic.com/v1` | Confirmed 200; US endpoint 403s with GOV key |
| Logs | `https://gov-log-api.newrelic.com/log/v1` | Per FedRAMP docs |
| Metrics | `https://gov-metric-api.newrelic.com/metric/v1` | Per FedRAMP docs |
| Synthetics | `https://synthetics.newrelic.com/synthetics/api` | No GOV variant exists (`gov-synthetics` has no DNS); falls back to US |
| InsightsKeys | `https://insights.newrelic.com/internal_api/1` | No GOV variant (`gov-insights` has no DNS); falls back to US |
| BlobService | `https://blob-api.service.newrelic.com/v1/e` | No GOV variant (`blob-api.service.gov` has no DNS); falls back to US |

All endpoint choices were verified by direct HTTP probe against the GOV test account (`4012030`) before being committed.

### Parse aliases

`Parse("gov")` and `Parse("fedramp")` both return `GOV` (case-insensitive), consistent with how `Parse("staging")` works.

### What's not included

- `gov-trace-api.newrelic.com` returns 202 and is a valid GOV endpoint, but `traceBaseURL` is not a field in the `Region` struct for any region — out of scope here.

## Test plan

- [x] `go test -tags unit ./pkg/region/...` — all 12 tests pass
- [x] GOV NerdGraph (`gov-api.newrelic.com/graphql`) confirmed 200 with GOV credentials
- [x] GOV insights-collector confirmed 200; US insights-collector confirmed 403 with same GOV license key
- [x] GOV OTLP (`gov-otlp.nr-data.net`) confirmed 200; US OTLP confirmed 403 with GOV license key
- [x] Infrastructure agent reporting to GOV endpoints validated end-to-end via `newrelic-cli` install on Ubuntu 22.04 EC2 (19 SystemSample events confirmed in GOV account via NRQL)

### DIFF
--- a/pkg/region/region_constants.go
+++ b/pkg/region/region_constants.go
@@ -15,6 +15,9 @@ const (
 	// JP represents New Relic's Japan-based production deployment.
 	JP Name = "JP"
 
+	// GOV represents New Relic's FedRAMP-compliant US government deployment.
+	GOV Name = "GOV"
+
 	// Staging represents New Relic's US-based staging deployment.
 	// This is for internal New Relic use only.
 	Staging Name = "Staging"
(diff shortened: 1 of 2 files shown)


nr-developer-toolkit (approved):

pranav-new-relic:
The failure with the E2M test is owing to an ongoing issue, unrelated to what we're doing. We should be good to proceed 👍

PR #1487, opened 2026-09-10
fix(cloud): distinguish absent from empty Azure Monitor tags

sgore-nr:
`CloudAzureMonitorIntegrationInput.IncludeTags` and `.ExcludeTags` carried `omitempty` on a plain slice, which cannot distinguish a nil slice from an empty one — `omitempty` drops both. Empty tags were therefore dropped from the mutation variables entirely, making it impossible to clear tags already configured on an Azure Monitor integration: sending no tags was indistinguishable from not mentioning them.

## Why not just remove `omitempty`

That was this PR's first approach, and it is wrong. Without `omitempty` a nil slice marshals to `null`, and NerdGraph rejects that outright:

```
ERR_INVALID_DATA include_tags must be a list
```

Verified against staging. Notably the API rejects `null` even though the field reads as nullable in the schema, so the resolver validates it explicitly rather than relying on GraphQL nullability.

That matters because nil is what callers produce for any `monitor` block that simply does not mention the tags — far more common than clearing them. So removing `omitempty` would have traded "`[]` silently does nothing" for "omitting the attributes errors out", breaking configurations that had always worked. A worse bug than the original.

## The fix

Make the fields pointers and keep `omitempty`. On a pointer, `omitempty` keys off the pointer rather than the length, so an empty slice behind a non-nil pointer still serializes:

```go
ExcludeTags *[]string `json:"excludeTags,omitempty"`
IncludeTags *[]string `json:"includeTags,omitempty"`
```

Three states, three distinct wire forms:

```
nil          -> field omitted        -> configured tags left unchanged
&[]string{}  -> "includeTags":[]     -> configured tags cleared
&[]string{x} -> "includeTags":["x"]  -> configured tags replaced
```

This mirrors `Enabled *bool` in the same struct, which became a pointer in #1485 for exactly this reason: `false` and "unspecified" are indistinguishable for a plain `bool` with `omitempty`.

Nothing about the API contract changes. JSON has no pointers, so a non-nil `*[]string` and a `[]string` serialize byte-for-byte identically; the GraphQL variable type is untouched. The pointer only buys the ability to *decline to emit a key* while still being able to emit `[]` — and omitting the key is behaviour the API already supported.

## Impact on callers

Callers that want to clear tags must pass a pointer to an empty slice. Passing nil keeps the previous "leave them alone" behaviour, so callers that never set these fields are unaffected at runtime, though the type change requires a compile-time update.

The Terraform provider does this from its raw configuration, where an absent attribute is null and `[]` is a known empty list: newrelic/terraform-provider-newrelic#3211.

## Verification

- `test-unit`, `test-integration`, `lint`, and `compile` on macOS/Ubuntu/Windows all pass.
- Exercised end to end against staging through the Terraform provider, for all three cases. The `null` variant is what surfaced the `ERR_INVALID_DATA` above; with pointers, omitting the attributes applies cleanly and setting `[]` clears the tags.

### DIFF
--- a/pkg/cloud/types.go
+++ b/pkg/cloud/types.go
@@ -2644,9 +2644,9 @@ type CloudAzureMonitorIntegrationInput struct {
 	// Specify if integration is active
 	Enabled bool `json:"enabled,omitempty"`
 	// Specify resource tags (in 'key:value' form) associated with the resources that you want to exclude from monitoring. Exclusion takes precedence over inclusion.
-	ExcludeTags []string `json:"excludeTags,omitempty"`
+	ExcludeTags []string `json:"excludeTags"`
 	// Specify resource tags (in 'key:value' form) associated with the resources that you want to monitor. If empty, all resources will be monitored.
-	IncludeTags []string `json:"includeTags,omitempty"`
+	IncludeTags []string `json:"includeTags"`
 	// The polling interval for metadata and tags, in seconds.
 	InventoryPollingInterval int `json:"inventoryPollingInterval,omitempty"`
 	// The linked account identifier.


pkg/cloud/types.go
```diff
@@ -2644,9 +2644,13 @@ type CloudAzureMonitorIntegrationInput struct {
 	// Specify if integration is active
 	Enabled *bool `json:"enabled,omitempty"`
 	// Specify resource tags (in 'key:value' form) associated with the resources that you want to exclude from monitoring. Exclusion takes precedence over inclusion.
-	ExcludeTags []string `json:"excludeTags,omitempty"`
+	// A nil pointer omits the field, leaving the configured tags unchanged; a pointer to an empty
+	// slice sends [], which clears them.
+	ExcludeTags *[]string `json:"excludeTags,omitempty"`
```
pranav-new-relic:
```suggestion
	// A nil pointer omits the field, leaving the configured tags unchanged; a pointer to an empty
	// slice sends [], which clears them.
	// This change has been made manually out of Tutone's purview, as a pointer override with primitive datatypes is not supported by Tutone.
	// DO NOT DISCARD THIS CHANGE when a refetch of this package is performed using Tutone.
	ExcludeTags *[]string `json:"excludeTags,omitempty"`
```

pkg/cloud/types.go
```diff
@@ -2644,9 +2644,13 @@ type CloudAzureMonitorIntegrationInput struct {
...
 	// Specify resource tags (in 'key:value' form) associated with the resources that you want to monitor. If empty, all resources will be monitored.
-	IncludeTags []string `json:"includeTags,omitempty"`
+	// A nil pointer omits the field, leaving the configured tags unchanged; a pointer to an empty
+	// slice sends [], which clears them.
+	IncludeTags *[]string `json:"includeTags,omitempty"`
```
pranav-new-relic:
```suggestion
	// A nil pointer omits the field, leaving the configured tags unchanged; a pointer to an empty
	// slice sends [], which clears them.
	// This change has been made manually out of Tutone's purview, as a pointer override with primitive datatypes is not supported by Tutone.
	// DO NOT DISCARD THIS CHANGE when a refetch of this package is performed using Tutone.
	IncludeTags *[]string `json:"includeTags,omitempty"`
```

PR #1360, opened 2026-01-13
feat(fleetcontrol): add new package with fleet, fleet members CRUD ops

pranav-new-relic:
### Overview
This branch introduces a comprehensive **Fleet Control** feature to the New Relic Go client library, enabling programmatic management of fleet deployments, agent configurations, and fleet member operations through the New Relic platform.

---

### 1. **Fleet Control Package** (`pkg/fleetcontrol`)
A new package providing complete CRUD operations for managing fleets of infrastructure entities.

#### Core Capabilities:
- **Fleet Management**
  - Create, update, and delete fleets
  - Add and remove fleet members
  - Query fleet members with filtering

- **Fleet Deployments**
  - Create, update, and delete fleet deployments
  - Execute deployment operations with `fleetControlDeploy`
  - Support for agent input configuration during deployments

- **Configuration Management (Blob Service Integration)**
  - Get agent configurations (active or draft versions)
  - Create and update agent configurations
  - Fetch specific configuration versions
  - Delete configurations and versions
  - List all configuration versions

#### Key Files:
- `fleetcontrol.go` - Core client initialization
- `fleetcontrol_api.go` - NerdGraph API mutations and queries (1,275 lines)
- `fleetcontrol_configurations.go` - Blob service integration for configurations (305 lines)
- `types.go` - Comprehensive type definitions (10,130 lines)
- Integration tests with 560 lines of coverage

---

### 2. **Tutone Configuration** (`.tutone.yml`)
Added complete package configuration for code generation:

**Mutations:**
- `fleetControlCreateFleet`
- `fleetControlUpdateFleet`
- `fleetControlDeleteFleet`
- `fleetControlAddFleetMembers`
- `fleetControlRemoveFleetMembers`
- `fleetControlCreateFleetDeployment`
- `fleetControlUpdateFleetDeployment`
- `fleetControlDeleteFleetDeployment`
- `fleetControlDeploy`

**Queries:**
- `actor.fleetControl.fleetMembers` - Query fleet members with filters
- `actor.entityManagement.entity` - Entity lookups
- `actor.entityManagement.entitySearch` - Entity search capabilities

**Type Overrides:**
- `FleetControlFleetMembersFilterInput` - Pointer type for optional filtering
- `FleetControlOperatingSystemCreateInput` - Optional OS specification
- Entity management types for fleets, configurations, and deployments

---

### 3. **HTTP Client Enhancements** (`internal/http/client.go`)
- Added support for custom HTTP methods (GET, POST, PUT, DELETE)
[...]

PR #1322, opened 2025-07-20
feat: support for custom headers routed via client-go

vinay-newrelic:
Screenshot 2025-07-21 at 1 42 07 AM
Screenshot 2025-07-21 at 1 42 19 AM

PR #1491, opened 2026-09-16
feat(http): add support for global custom HTTP headers via Config

Nandu-pns:
## Summary

Adds a `Config.CustomHeaders` option so callers can configure HTTP headers that are automatically applied to **every** outgoing request made by the client — the "global config for persistent headers" option proposed in [NR-436498](https://new-relic.atlassian.net/browse/NR-436498), primarily to support ELB rate-limiting headers (e.g. `X-Client-ID`) and NerdGraph routing headers.

Closes [NR-436600](https://new-relic.atlassian.net/browse/NR-436600) / [NR-436498](https://new-relic.atlassian.net/browse/NR-436498).

## Why not build on #1322?

NR-436600 asked for a review of #1322 as a path to close this out. I did a full review + local repro, and found real issues that make it unsuitable to build on directly:
- Bundles in an unrelated dashboards change (`pkg/dashboards/types.go`, multi-account-ID support for widgets — nothing to do with headers).
- Adds a `Client.customHeaders` field that's written via `SetCustomHeaders` but **never read** — `makeRequest()` actually reads `r.config.CustomHeaders`, so that method is dead code that looks functional but isn't.
- Leaves the original `PostWithContext` body commented out instead of removed.
- Only wires headers into POST + NerdGraph, not Get/Put/Delete.
- Currently `CONFLICTING` with `main`, and the CLA is unsigned.

Separately, I found `main` already has a narrower custom-headers mechanism (`getRequestCustomHeadersFromQueryParams`/`setRequestCustomHeaders` in `internal/http/client.go`, added in #1360), but it's `pkg/fleetcontrol`-specific plumbing (confirmed via grep — that's its only caller anywhere in the repo), wired only into `PostWithContext` via a sentinel key smuggled into `queryParams`, and it discards any other query params when that key is present. Left untouched here — out of scope for this change, but worth a cleanup follow-up someday.

## What this PR actually does

- `pkg/config/config.go`: adds `Config.CustomHeaders map[string]string`.
- `pkg/config/main.go`: adds `ConfigCustomHeaders(headers map[string]string) ConfigOption`, following the exact pattern of the other options already in this file. Merges into any existing headers rather than replacing.
- `newrelic/newrelic.go`: adds the top-level `ConfigCustomHeaders` wrapper, matching `ConfigLogger`/`ConfigUserAgent`.
- `internal/http/request.go`: in `NewRequest`, applies `cfg.CustomHeaders` **before** the SDK's own reserved headers (`NewRelic-Requesting-Services`, `Content-Type`, `User-Agent`) — so those defaults always win on an accidental name collision, and custom headers can't clobber internal service-name tracking.

Since every verb (`Get`, `Post`, `Put`, `Delete`, `NerdGraphQuery`) already routes through the same `NewRequest`, this one change covers all of them uniformly — no per-verb duplication needed.

**Per-request overrides need no new code.** `Request.SetHeader`/`GetHeader`/`DelHeader` already exist and are already public — any `req.SetHeader(...)` call after `NewRequest` returns overwrites whatever was set from `Config.CustomHeaders`, exactly matching the "per-request override" half of the original ask. I added a test (`TestConfigCustomHeadersOverriddenByRequest`) proving this rather than assuming it.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `gofmt -l` / `goimports -l` on changed files — clean
- [x] `golangci-lint run` on changed packages — 0 issues
- [x] `misspell` on changed files — clean
- [x] `go test -tags unit ./...` — full suite passes; new tests (`TestConfigCustomHeaders`, `TestConfigCustomHeadersOverriddenByRequest`, `TestConfigCustomHeadersAppliedToNerdGraphRequest`) pass, no existing tests regressed. The only failures in the full run (`pkg/customeradministration`, `pkg/organization`) are pre-existing `TestIntegration*` tests requiring live New Relic credentials not available in a local sandbox — confirmed by running the identical suite against unmodified `main`, which fails the same way.
- [x] Confirmed no changes to `pkg/dashboards` or any auth/credential-handling code path; diff scanned for secrets, none found.

## Backward compatibility

Specifically checked this doesn't break the Terraform provider or any other direct client-go consumer, since every Terraform request goes through this exact code path:

- **Every existing caller has `Config.CustomHeaders == nil`** (it's a brand-new field). Ranging over a nil map in Go is a proven no-op — verified with a standalone repro, not assumed — so the new loop in `NewRequest` executes zero times for anyone not using the new option. The request that goes out on the wire is byte-for-byte identical to before this PR for all existing callers.
- **No existing exported signature, field, or behavior was changed** — confirmed via `git diff main...HEAD`, every change is a pure addition (`+`), zero `-func`/`-field` removals.
- **Checked terraform-provider-newrelic's actual code on GitHub**: it constructs `config.Config{...}` via keyed struct literals with only 1-2 fields set (e.g. `config.Config{PersonalAPIKey: testAccAPIKey}`), and its production path uses the `ConfigOption` pattern (`newrelic.New(newrelic.ConfigPersonalAPIKey(...))`) — both are safe against a new field/new option being added; keyed literals never break on struct field additions, and `ConfigOption` is additive by design.
- **Checked `newrelic-cli`** (the other major direct consumer) — same `newrelic.New(cfgOpts...)` pattern, unaffected.
- Ruled out `exhaustruct`-style linters (which could flag a new struct field in CI) — not configured in this repo or terraform-provider-newrelic.
- No reflection or JSON/serialization of `Config` anywhere in this codebase that could expose the new field unexpectedly.
- `test-integration` in this PR's own CI already exercises the full existing test suite against real New Relic APIs with real credentials, through this same code path, and passes — the most realistic confirmation available that existing request behavior is unchanged.

**Net: this is additive-only and fully gated behind explicitly opting into `ConfigCustomHeaders`. Nobody gets a new header, changed behavior, or a different request on the wire unless they deliberately use the new option.**

[NR-436498]: https://new-relic.atlassian.net/browse/NR-436498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NR-436600]: https://new-relic.atlassian.net/browse/NR-436600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NR-436498]: https://new-relic.atlassian.net/browse/NR-436498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Nandu-pns:
Re: the Codecov patch-coverage flag on `pkg/config/main.go` — every line this PR adds is fully exercised by tests; the 0% Codecov reports there is a pre-existing measurement artifact of this package, not a real gap.

- `newrelic/newrelic.go`'s `ConfigCustomHeaders` wrapper: **100%** per Codecov's own report.
- `pkg/config/main.go`'s `ConfigCustomHeaders`: verified **100%** covered by the same tests (`go test -coverpkg=./...` shows every line executed) — it only reports 0% under this repo's actual CI invocation (`build/test.mk`, no `-coverpkg`), because `pkg/config` has no test files of its own, so cross-package calls into it (from `newrelic_test.go`, in this case) can't be attributed back to its coverage number.
- Confirmed this isn't specific to this PR: all 15 *pre-existing* `ConfigOption` functions in the same file (`ConfigPersonalAPIKey`, `ConfigLogger`, `ConfigRegion`, etc.) show the identical 0% under this repo's measurement setup, despite being exercised the same way via `newrelic_test.go`. Fixing it would mean adding a first-ever test file to `pkg/config` for every option in the file — out of scope here.

